PR #15, opened 2018-02-01
Create Botam

shazdeh021p:


levlam:
I don't think that translation of Bot API answerInlineQuery method to Persian needs to be added to the TDLib examples.

PR #21, opened 2018-02-01
Minimal golang binding with tdjson

cosmonawt:


MetalBlueberry:
I think this example is too simple,
take a look to L11R example, I've found it really useful
https://github.com/L11R/go-tdjson

cosmonawt:
It's just to illustrate how to use CGo to create the binding. I am aware of @L11R's and that it is more practical.

PR #117, opened 2018-03-02
Fix build of tdactor example

danog:
Then cmake file specified that a target with name `example` when building a file from a directory with the same name.
This triggers an error when building td as a submodule included with add_subdirectory.
Other fatal errors are also triggered when building like this, mainly complaining about non-existent headers due to the absence of tdutils and td/generate/auto from the include path, I fixed them by including these paths manually in my cmake file, am I missing anything?


levlam:
Could you show your CMakeLists.txt and how you run cmake?

danog:
```
cmake_minimum_required(VERSION 3.1 FATAL_ERROR)

project(PhpTd VERSION 1.0 LANGUAGES CXX)

add_subdirectory(td)
include_directories(td/td/generate/auto)
include_directories(td/tdutils)

add_library(phptd SHARED main.cpp)
target_link_libraries(phptd PRIVATE Td::TdJsonStatic phpcpp)
set_property(TARGET phptd PROPERTY CXX_STANDARD 14)
```
I then run 
```
cmake -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo .
cmake --build .
```

levlam:
The problem with include_directories is known and will be fixed in the next release.

To mitigate the second error out-of-source build should be used: https://cmake.org/Wiki/CMake_FAQ#Out-of-source_build_trees.
In-source builds have a lot of drawbacks, so we don't want to encourage they usage.

danog:
Oh ok then.
Should I close this PR? Having a cmake file with matching source and target name is still kinda ugly...

levlam:
The problem isn't in matching source and target names. That is nevertheless a good naming. The problem is that when in-source build is used CMAKE_CURRENT_SOURCE_DIR coincides with CMAKE_CURRENT_BINARY_DIR, so build create a lot of files in the project source tree. It can occasionally rewrite some of the source files or break everything in some other unpredictable way.
Herein the problem arises because path to executable is "<CMAKE_CURRENT_BINARY_DIR>/tdactor/example" matches name of existing directory "<CMAKE_CURRENT_SOURCE_DIR>/tdactor/example". And it is very good that it can't rewrite directory with an ordinary file and fails fast.

I prefer this not to be fixed, but prohibition of in-source builds should be more explicit.

levlam:
Between, what headers was included in main.cpp, so additional include_directories became mandatory? Now it looks like there is some missed dependency of phptd target.

levlam:
In-source builds were explicitly prohibited in TDLib 1.2.0, so I'm closing this.

Issue #141, opened 2018-03-12
Java example not working properly in v. 1.1.1 (in non-cmd consoles) [solved]


hekkup:
When you build and run the Java example in v. 1.1.1, it just prints some JSON stuff and then stops doing anything.

With some debugging I figured out the program keeps waiting for authorization in main().

By enabling the print() function in Client.UpdatesHandler.onResult() default case, the handler prints out many more JSON messages. Looks like message parsing/handling is out of date somewhere. Continuing digging...

levlam:
Hi.
Are you running the example on Windows? Are you using cmd as a console?

hekkup:
Ok, the example works in cmd on Windows.

It doesn't work in Git Bash (MingW64 based).

Java version is 1.8.0-161.

I'm running the program with command
```
 java -d32 -Djava.library.path=. org/drinkless/tdlib/example/Example
```

So the program runs in 32-bit mode. It's because compiling TdLib in 64-bit mode was not successful out of the box so I just went with 32-bit build. There were some problems with CMake settings (I built it as instructions said, though).

Could this issue with cmd vs other consoles be added into the instructions, please?

And is there some remedy for that?

levlam:
Yes, I'll add a note that on Windows the example should be run from cmd. Git bash uses mintty and mintty is incompatible with Java's System.console() used in the example. Some details about why it can't be fixed on mintty side can be found here: https://github.com/mintty/mintty/issues/244. Other similar issues with mintty on Windows can be found here: https://github.com/mintty/mintty/issues/56.

To built 64-bit binary on Windows, you need to specify -A x64 to CMake. This possibly also needs to be added to README, despite it is entirely specific to CMake's MSVC generator.

juananpe:
Hi there. Same problem here, but in this case, running the example from Eclipse's console. I can build and run the example, but it only goes up to the test of Client.execute (I think that that's the JSON output referred  by @hekkup) . Then the code enters the main loop and gets blocked. I have found that it is due the System.console() command. You can't run it inside Eclipse's console, but there are some alternatives: https://stackoverflow.com/questions/104254/java-io-console-support-in-eclipse-ide

I've fixed it substituting the System.console().readline() line by this snippet: 

	```
System.out.println(
				"Enter command (gcs - GetChats, gc <chatId> - GetChat, me - GetMe, sm <chatId> <message> - SendMessage, lo - LogOut, q - Quit): ");
[...]

PR #146, opened 2018-03-15
Fixed non-cmd console issue in Java example + added more build info

hekkup:
Fixes https://github.com/tdlib/td/issues/141

### DIFF
--- a/example/java/README.md
+++ b/example/java/README.md
@@ -8,10 +8,12 @@ TDLib should be prebuilt for using with Java and installed to local subdirectory
 cd <path to TDLib sources>
 mkdir jnibuild
 cd jnibuild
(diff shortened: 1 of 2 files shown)


levlam:
Hi, @hekkup.
Thanks for your contribution.
Java example was changed slightly in 1.2.0, so could you
1. Rebase your commit on top of master.
2. Remove also new mentions of System.console added in 1.2.0. There was a warning added:
```
        if (console == null) {
            System.err.println("This example requires Console for user interaction. On Windows cmd or PowerShell should be used");
            System.exit(1);
        }
```
and a variable `console` was added.
3. Do not add `-DCMAKE_TOOLCHAIN_FILE=<VCPKG_DIR>/scripts/buildsystems/vcpkg.cmake ` to
```
cmake -DCMAKE_BUILD_TYPE=Release -DTD_ENABLE_JNI=ON -DCMAKE_INSTALL_PREFIX:PATH=../example/java/td ..
```
because this option is specific to Windows and vcpkg and can't be used in general purpose instructions. It also shouldn't be needed if vcpkg is integrated with Visual Studio.
4. Could you also remove phrase "More info about 64-bit build over here: https://github.com/tdlib/td/issues/20", because `-A x64` should be really enough for 64-bit build and it is hard to find useful information about 64-bit build in comments to that issue.

Then I'm going to merge the fix to master.

PR #153, opened 2018-03-20
Update README.md

hostmaster:
c++ compiler runs out of memory on a 1Gb RAM VPS. More swap space needs to be added.


isopen:
This is more relevant to setting up VPS for assembly. It is better to write this to the provider of the VPS.

PR #246, opened 2018-05-18
Clarify the message id description (fixes #245)

### LINKED ISSUE #245: Message ids aren't unique
[The API docs states](https://core.telegram.org/tdlib/docs/classtd_1_1td__api_1_1message.html) that message ids are unique:

cloudapp annotation 2018-05-18 at 11 09 54 am png

Although I just discovered that it's not true. Hence the question: how are they work? Are ids are unique to a chat or combination of a chat and a client? If I'd get the same message on a different client (an instance of TDLib) will message have the same id, or it will be completely different?

Here is an example of an incoming update:

cloudapp annotation 2018-05-18 at 10 59 28 am png

And here is a message stored in my DB:

cloudapp annotation 2018-05-18 at 11 08 24 am png

This is really confusing. Thank you in advance!

kossnocorp:
Clarify that the message id is unique to the chat it belongs to.

See https://github.com/tdlib/td/issues/245

### DIFF
--- a/td/generate/scheme/td_api.tl
+++ b/td/generate/scheme/td_api.tl
@@ -450,7 +450,7 @@ messageSendingStateFailed = MessageSendingState;
 
 
 //@description Describes a message
-//@id Unique message identifier
+//@id Message identifier unique to the chat it belongs to.
 //@sender_user_id Identifier of the user who sent the message; 0 if unknown. It is unknown for channel posts
 //@chat_id Chat identifier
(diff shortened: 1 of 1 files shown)


UnblockerUsername:
Cool. Thus, you narrow the scale of uniqueness.

UnblockerUsername:
@levlam How about to narrow the scale of uniqueness?

PR #274, opened 2018-06-05
Add minimal javascript example

Bannerets:


wfjsw:
I would prefer node C/C++ addon over ffi :/

PR #281, opened 2018-06-10
Make result type safe

mincong-h:
I'd like to add a generic type T to `Client.ResultHandler`, which helps to determine the result type and avoid unnecessary switch-case statement. For example, querying the chat history using `TdApi#GetChatHistory(...)` will return a `TdApi.Messages` object.

Previously, we need to do:

```java
private static class MessageHandler implements Client.ResultHandler {
  @Override
  public void onResult(TdApi.Object object) {
    switch (object.getConstructor()) {
      case Messages.CONSTRUCTOR: {
        Messages messages = (Messages) object;
        for (Message msg : messages.messages) {
          ...
        }
        break;
      }
      default:
        System.out.println("This should never happen");
        break;
    }
  }
}
```

Now we can do:

```java
private static class MessageHandler implements Client.ResultHandler<Messages> {
  @Override
  public void onResult(Messages messages) {
    for (Message msg : messages.messages) {
      ...
    }
  }
}
```

### DIFF
--- a/example/java/org/drinkless/tdlib/Client.java
+++ b/example/java/org/drinkless/tdlib/Client.java
@@ -18,14 +18,16 @@ import java.util.concurrent.locks.ReentrantReadWriteLock;
 public final class Client implements Runnable {
     /**
      * Interface for handler for results of queries to TDLib and incoming updates from TDLib.
+     *
+     * @param <T> The result type
      */
-    public interface ResultHandler {
+    public interface ResultHandler<T extends TdApi.Object> {
         /**
          * Callback called on result of query to TDLib or incoming update from TDLib.
          *
          * @param object Result of query or update of type TdApi.Update about new events.
          */
-        void onResult(TdApi.Object object);
+        void onResult(T object);
     }
 
     /**


levlam:
You suggest a good change, but assumption
```
System.out.println("This should never happen");
```
is wrong, because every method can return TdApi.Error instead of a specified object.
This can be solved by providing methods `onSuccess(T result)` and `onError(TdApi.Error error)`, but it is not easy to do without breaking existing code. In case of these changes a programmer also need to choose, whether he wants to manually check result type or he wants to write two functions.

mincong-h:
> You suggest a good change, but assumption
>
>     System.out.println("This should never happen");
>
>is wrong, because every method can return TdApi.Error instead of a specified object.

I thought the exception handled by `Client.ExceptionHandler`, because method `Client#send` accept explicitly an exception handler:

```java
public void send(TdApi.Function query, ResultHandler resultHandler, ExceptionHandler exceptionHandler) {
    ...
}
```

But you're right. The exception handler is present only for handling cases when Java fails the deserialize the result. If deserialization goes well, there will always be 2 cases—success and failure, both of them are present as `TdApi.Object`.

> This can be solved by providing methods `onSuccess(T result)` and `onError(TdApi.Error error)`, but it is not easy to do without breaking existing code. In case of these changes a programmer also need to choose, whether he wants to manually check result type or he wants to write two functions.

I agree. Let's keep it as it is.

PR #287, opened 2018-06-14
Fixed ARM UWP Build with Visual Studio 2017

thunderluca:
With Visual Studio 2017 build tools/IDE, the current build.ps1 script always throws "error MSB8022: Building Desktop applications for the ARM platform is not supported" every time that cmake tries to generate project files for arm architecture.

Build tools seem to not recognize the "arm" architecture or they are trying to build non-UWP projects because of lower case architecture name (with the same tools and SDK, the build configuration only succed when "arm" is written in upper case).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,3 @@ auto/
 db_backup
 *.pyc
 docs/
-/.vs/td/v15/Browse.VC.opendb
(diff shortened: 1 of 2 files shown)


levlam:
Thanks, @thunderluca!

Richiedevop (approved):

PR #303, opened 2018-07-02
Use standard escape sequences

mbasaglia:
In tdutils/td/utils/logging.h it had the non-standard \e escape, which causes some warnings when compiling in standard mode (or errors with -Werror) so I replaced it with the standard hex escape sequence.


levlam:
Thanks, @mbasaglia.

PR #314, opened 2018-07-19
Add a function to get the current log verbosity

mbasaglia:
I noticed that Log has a function to set log verbosity but not to get the current value, so I've added a function that does that.

The rationale is that I wanted to write a client using tdlib where the logging level is configurable, so I need a way to retrieve it.

### DIFF
--- a/td/telegram/Log.cpp
+++ b/td/telegram/Log.cpp
@@ -46,6 +46,10 @@ void Log::set_verbosity_level(int new_verbosity_level) {
   SET_VERBOSITY_LEVEL(VERBOSITY_NAME(FATAL) + new_verbosity_level);
 }
 
+int Log::get_verbosity_level() {
+  return GET_VERBOSITY_LEVEL();
+}
+
 void Log::set_fatal_error_callback(FatalErrorCallbackPtr callback) {
   if (callback == nullptr) {
     fatal_error_callback = nullptr;
--- a/td/telegram/Log.h
+++ b/td/telegram/Log.h
@@ -59,6 +59,13 @@ class Log {
   static void set_verbosity_level(int new_verbosity_level);
 
   /**
+   * Returns the verbosity level of the internal logging of TDLib.
+   *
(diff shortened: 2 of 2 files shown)


levlam:
The log verbosity level isn't saved between application restarts, so this method is needed only if you don't want to save on your side last parameter of Log::set_verbosity_level, Is it the case?

mbasaglia:
Yes,  I want to implement the logic to save and restore the verbosity level.

levlam:
Where do you want verbosity level to be saved? Should it be persistent between application restarts for your needs?

mbasaglia:
I'm implementing log verbosity persistence myself, hence this pull request to retrieve the log level. This allows me to save it when the application shuts down.

levlam:
@mbasaglia Your motivation for adding this functionality to TDLib interface is perfectly valid, but I'm going to reject the pull request because
1) TDLib has JSON (C) and native Java and C# interfaces, which should be changed accordingly if this method is added;
2) Any change in TDLib external interface needs to be supported in all TDLib-based framework;
3) I don't want to further expand log managing part of TDLib interface, because otherwise it will have more methods than main part.
4) The same behavior can be achieved by simple wrapper to `Log::set_verbosity_level`:
```
class TdlibLog {
  static int verbosity_level_ = 5;
 public:

  static void set_verbosity_level(int new_verbosity_level) {
    verbosity_level_ = new_verbosity_level:
    td::Log::set_verbosity_level(new_verbosity_level)
  }

  static int get_verbosity_level() {
    return verbosity_level_;
  }
};
```

mbasaglia:
It just seems very weird to have a setter without a getter, especially since the functionality already exists internally

levlam:
@mbasaglia In TDLib 1.4.0 the API for managing the TDLib internal logging was revorked. Now the logging can be managed through ordinary TDLib methods. In particular, the synchronous method `getLogVerbosityLevel` was added, which can be used to get the current verbosity level of logging.

PR #359, opened 2018-09-11
Add events for signin process

tallero:
Added the remaining steps to complete a signin.

### DIFF
--- a/example/python/tdjson_example.py
+++ b/example/python/tdjson_example.py
@@ -1,5 +1,6 @@
 #
-# Copyright Aliaksei Levin (levlam@telegram.org), Arseny Smirnov (arseny30@gmail.com) 2014-2018
+# Copyright Aliaksei Levin (levlam@telegram.org), Arseny Smirnov (arseny30@gmail.com),
+# Pellegrino Prevete (pellegrinoprevete@gmail.com)  2014-2018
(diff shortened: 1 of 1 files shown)


example/python/tdjson_example.py
```diff
@@ -1,5 +1,6 @@
 #
-# Copyright Aliaksei Levin (levlam@telegram.org), Arseny Smirnov (arseny30@gmail.com) 2014-2018
+# Copyright Aliaksei Levin (levlam@telegram.org), Arseny Smirnov (arseny30@gmail.com),
+# Pellegrino Prevete (pellegrinoprevete@gmail.com)  2014-2018
```
kossnocorp: I'm representing neither Telegram nor the TDLib team, but I believe that contributors couldn't claim copyright by contributing to the project.

tallero: I'm not aware of CTAs for this project; also there should not be problems about relicensing (in a remote future) since the Boost license permits proprietary derivatives.
Finally, I would only get copyright for this file.

I said to myself 'why not? If I have to contribute to something anyone could relicense as proprietary, at least I can let it bring my name'.

Reference about contributors in source code:
https://opensource.stackexchange.com/a/888

PR #362, opened 2018-09-17
Added SOVERSION to shared libraries

xvitaly:
Added SOVERSION to shared libraries.
SOVERSION required for all major GNU/Linux distributions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -804,3 +804,7 @@ write_basic_package_version_file("TdConfigVersion.cmake"
 install(FILES "TdConfig.cmake" "${CMAKE_CURRENT_BINARY_DIR}/TdConfigVersion.cmake"
   DESTINATION lib/cmake/Td
 )
+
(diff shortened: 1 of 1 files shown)


xvitaly:
> Despite different TDLib versions are tend to be ABI compatible, underlying JSON API is different, so there is no compatibility guarantee between different minor versions

Very bad. Please read about [SEMVER](https://semver.org/).

> So shouldn't we use full TDLib version as SOVERSION instead?

No problem. I can update my PR to use full version.

xvitaly:
Updated PR to include project version as SOVERSION.

levlam:
> Very bad. Please read about SEMVER.
TDLib JSON interface adheres to SEMVER semantics and the same client code can work with any version of libtdjson.
In contrast, TDLib native C++ interface tdclient is strongly typed and can't adhere to SEMVER, because otherwise current TDLib version would be something like 97.0.1.

PR #363, opened 2018-09-17
Use system-wide crypto policies

xvitaly:
Use system-wide default crypto policies. More information: https://fedoraproject.org/wiki/Packaging:CryptoPolicies

### DIFF
--- a/tdnet/td/net/SslStream.cpp
+++ b/tdnet/td/net/SslStream.cpp
@@ -295,7 +295,7 @@ class SslStreamImpl {
 
     // TODO(now): cipher list
     string cipher_list;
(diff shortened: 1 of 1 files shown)


xvitaly:
Yes, [Fedora specific](https://src.fedoraproject.org/rpms/openssl/blob/master/f/openssl-1.1.1-system-cipherlist.patch). Closing this PR.

levlam:
Between, is there a recommended way to detect in build time that TDLib is built for Fedora and change the cipher list only in that case?

xvitaly:
> Between, is there a recommended way to detect in build time that TDLib is built for Fedora and change the cipher list only in that case?

Fedora will have [native tdlib package](https://src.fedoraproject.org/rpms/tdlib) soon. We added patch.

PR #435, opened 2018-12-16
lua example

sajjad-021:
add example for lua

### DIFF
--- a/example/lua/README.md
+++ b/example/lua/README.md
@@ -1,0 +1,10 @@
+# TDLib Lua example
+
+### TDLua
+**basic wrapper for tdlib**
+
+#### Requirements
+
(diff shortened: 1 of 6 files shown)


levlam:
Hi, @sajjad-021. Thanks for you contribution.
I think that we don't need more examples of TDLib integration with different programming languages in the main repository. They will be better handled and kept up-to-date in third-party repositories, links to which will be added to README,md.

These Lua examples are already available in two repositories listed in our [README](https://github.com/tdlib/td/tree/master/example#using-tdlib-in-lua-projects), so there is no reason to include them directly.

sajjad-021:
thanks for your response.
Since you chosen the my source for lua example,
I was think that better add it to library.
thank so much.

PR #471, opened 2019-01-26
Preliminary FreeBSD support

vishwin:
- Add `__FreeBSD__` define
- FreeBSD and Darwin use the same poll implementation
- Include a needed header for `sockaddr_in{,6}`

Builds successfully on FreeBSD 13-CURRENT; should have no problem with 12.0-RELEASE, 12-STABLE, 11.2-RELEASE or 11-STABLE.

### DIFF
--- a/tdutils/td/utils/port/IPAddress.h
+++ b/tdutils/td/utils/port/IPAddress.h
@@ -18,6 +18,10 @@
 #include <sys/socket.h>
 #endif
 
+#if TD_FREEBSD
+#include <netinet/in.h>
+#endif
+
 namespace td {
 
 Result<string> idn_to_ascii(CSlice host);
--- a/tdutils/td/utils/port/config.h
+++ b/tdutils/td/utils/port/config.h
@@ -25,7 +25,7 @@
 #elif TD_EMSCRIPTEN
   #define TD_POLL_POLL 1
   #define TD_EVENTFD_UNSUPPORTED 1
-#elif TD_DARWIN
+#elif TD_DARWIN || TD_FREEBSD
   #define TD_POLL_KQUEUE 1
   #define TD_EVENTFD_BSD 1
 #elif TD_WINDOWS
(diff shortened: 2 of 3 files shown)


levlam:
FreeBSD out-of-the-box support was already added in our private unreleased version, which will be released soon, but thanks for the patch anyway. In meanwhile it can be used to build TDLib 1.3.0 on FreeBSD.

vishwin:
Yes, building 1.3.0 was the point of this patch. Looking forward to the next version!

omar-polo:
Hi,

I'm just a random guy on the internet who was going to submit a pr similar to this to add OpenBSD support. (But the following should apply to NetBSD and other unices as well).

Following what [POSIX says](http://pubs.opengroup.org/onlinepubs/000095399/basedefs/netinet/in.h.html) (if I'm not misunderstanding) `sockaddr_in` should be defined in `netinet/in.h`. I suppose that on linux and/or macos that header is pulled in by some other header.

So, in my humble opinion, `netinet/in.h` should always be included (at least on unix) and not only for FreeBSD.

levlam:
@omar-polo This is true, we will have `netinet/in.h` inlcuded on all POSIX-compliant platforms, including OpenBSD and NetBSD.

levlam:
Hi @vishwin and @omar-polo,

FreeBSD out of the box support is available in TDLib 1.4.0. You can also choose it as an operating system in a [TDLib build instructions generator](https://tdlib.github.io/td/build.html).

OpenBSD and NetBSD support will be added later. This can likely be done in a few lines of code, but they will also need to be supported in the build instructions generator and thoroughly tested for compatibility.

levlam:
@omar-polo OpenBSD and NetBSD out of the box support is available in TDLib 1.5.0. They also can be chosen as an operating system in a [TDLib build instructions generator](https://tdlib.github.io/td/build.html).

omar-polo:
Thanks for your work @levlam! I'll test the latest version soon!

I gave up with the pr because I was getting an error when trying
to run the tests iirc (albeit the library was working.)

Issue #372, opened 2018-10-01
Can't build iOS library


istarchenkov:
Get error when running build.sh

./build.sh: line 2: realpath: command not found
Platform = macOS Simulator = 
./build.sh: line 12: realpath: command not found
OpenSSL path = 
CMake Error: The source directory "/Development/TDLib/example/ios/build/build-macOS" does not appear to contain CMakeLists.txt.
Specify --help for usage, or press the help button on the CMake GUI.
make: *** No rule to make target `install'.  Stop.

jwktje:
I can't get it to work. I get these errors every time. 
https://pastebin.com/rj2BmA3z

Any idea @levlam ?

levlam:
@jwktje There is a problem with OpenSSL build. Could you provide output of `./build-openssl.sh` and file `build-ios/CMakeCache.txt`?

jwktje:
Sure no problem! The sh output is after I've ran it successfully before. Is that enough info or do you need the output from an actual fresh build? 
[./build-openssl.sh output Pastebin](https://pastebin.com/jVstxTNx)
[build-ios/CMakeCache.txt Pastebin](https://pastebin.com/0yWKmKmp)

jwktje:
Did exactly that just now, and this is the full output. Error seems the same as my initial pastebin. 

https://pastebin.com/dK3pPEKE

levlam:
@jwktje Thanks, for the info. It seems that the problem is in our usage of `Python-Apple-support`, which dropped support for armv7 build of OpenSSL in https://github.com/pybee/Python-Apple-support/commit/ec1706d67ed3a01f80df53ac22d004db94a0812a. Could you replace
`git checkout 2.7`
in `build-openssl.sh` with
`git checkout 60b990128d5f1f04c336ff66594574515ab56604` and rerun it?

jwktje:
That seems to work! It builds and I get the header files. Can't find the .a files afterwards though. Where should they end up?

Khwuanchai:
I​ need​ it.

ในวันที่ อา. 2 ธ.ค. 2018 07:27 น. Aliaksei Levin <notifications@github.com>
เขียนว่า:

AmirCa84:
[...]

PR #502, opened 2019-03-03
Fix Python-Apple-Support dependency checkout

ERussel:
- Last Python-Support release dropped support for arm7, armv7s builds so previous release tag must be pined until corresponding archs will be dropped from tdlib.

Signed-off-by: Russel <emkil.russel@gmail.com>

### DIFF
--- a/example/ios/build-openssl.sh
+++ b/example/ios/build-openssl.sh
@@ -2,7 +2,7 @@
 
 git clone https://github.com/pybee/Python-Apple-support
 cd Python-Apple-support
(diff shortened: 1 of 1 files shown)


levlam:
Thanks, @ERussel! We have similar fix in our private repository, so the pull request will not be merged, but I've mentioned it in the corresponding issue https://github.com/tdlib/td/issues/372.

levlam:
The issue was fixed in TDLib 1.4.0 in a similar manner.

PR #604, opened 2019-06-22
[Bugfix][Example][Web] Fix typo after import

kesha-antonov:
Hey!

Working on telegram-react
Encountered this issue https://github.com/evgeny-nadymov/telegram-react/issues/33
And then tried ```mode: 'asmjs'``` and stumbled upon this call.

PR #636, opened 2019-07-23
Fix for loading asm.js version of tdweb

dstaley:
When attempting to run the asmjs version of tdweb, I ran into a few issues. This PR fixes them:

1. `log.warning` is replaced with the correct call, `log.warn`
1. `Module.default` is used as opposed to `Module`


levlam:
Thanks, @dstaley. The pull request will be merged when the next TDLib version is released.

The second fix is the same as in https://github.com/tdlib/td/pull/604.

isopen:
@dstaley It was sumptuously. I will use these types of pull requests :+1:

PR #637, opened 2019-07-25
Replaced with a better library

AmirSbss:
Hi, this lib is based on the old one but it's much better and well documented


levlam:
I see some good changes in your fork, but I don't see a reason to remove link to the original actively developed library from the list of examples.

levlam:
@AmirSbss I've added a link to your fork to the list of examples in https://github.com/tdlib/td/commit/449bd82fad3fbd5b351f219cb7b8dfbaa15805f9.

PR #678, opened 2019-08-28
Fix buffered reader typo

d1musk:


levlam:
Thanks, @d1musk! This misprint was already fixed in our private version, so it will be fixed in TDLib 1.5.0.

levlam:
Closing the pull request because our own fix of the misprint is already public.

PR #711, opened 2019-09-29
 ^tab

isopen:


ratijas:
Hello. I believe, sqlite3 sources are autogenerated/assembles at upstream from millions of their source files, and are not meant to be edited afterwards. Why would you go through and change tabs vs spaces here instead of resolving the issue with sqlite3 itself?

isopen:
Everything is obvious. So that in subsequent associations there are no extra characters.

levlam:
@isopen The SQLCipher amalgamation is automatically generated from its source code and is updated from time to time, so it should have as small patches as possible. There is no way we could maintain whitespaces changes in the amalgamation and the change doesn’t affect anything, so we have no reason to apply it.

isopen:
@levlam This is clear. After assembling in one piece, you can do a cleanup of the code from extra characters to avoid such commits.

ratijas:
Well, that was kinda fun (in a good sense). Don't get me wrong, I absolutely LOVE cosmits (cosmetic commits) where appropriate. And your passion for it is what makes _you_ the professional, @isopen. Just -- you know -- the [appropriate](https://www.sqlite.org/amalgamation.html) [places](https://sqlite.org/src/timeline)... I think you could fix it upstream and make all the developer-users of sqlite around the world a little bit happier. They have very [strict](https://sqlite.org/copyright.html) contribution policy, but otherwise are opened to suggestions.

Anyway, @levlam, is that referenced commit relevant anymore? Because even that define+include looks like it could be extracted into .patch or `sed` script to be applied at compile time.

levlam:
@ratijas It is a bad idea to patch commited source files during build phase.

isopen:
@levlam Maybe you should change the storage to a more reliable one?

ratijas:
Well, the Arch Linux community and whole *NIX packaging system is a bad idea then :)

But I see your point, and I I see how easily things could go wrong there. I can't think of any other alternative, and so I have nothing more to say on topic. Have a nice day!

@isopen In order to be productive, one should appeal to the facts instead of making propositions in a form of "maybe you should ...?" questions. Please, describe what makes you think that sqlite3 (which is used in literally every coffee machine around the globe) is unreliable and what are the proposed alternatives?

isopen:
I spent some time reading the sqlite3 code, and this is just some kind of duck. You can see this also by spending some time and reading the sqlite3 code. For example, Preston might start writing not only for QT.

levlam:
So, the first commit can't be accepted, because the changes can't be maintened in the future.
The second commit deletes a comment and a TODO, and that's a wrong way to handle the TODO.
The third commit adds `.idea` to `.gitignore`. This change doesn't harm and can help someone, but the change is coming with two other commits, so there is no way we can merge it,

ratijas:
Also, .idea, *.sublime-settings, *.vim~, .directory, .DS_Dtore etc. etc. are more useful when placed in global user config like `~/.gitignore` (however, does not hurt in repository when the whole team uses the same toolchain — which is not the case for most of GitHub).

isopen:
@levlam https://en.wikipedia.org/wiki/Lernaean_Hydra

PR #716, opened 2019-10-02
tdutils: Fix build for alpine linux

Akianonymus:
* Alpine doesn't use sys/sysctl header for syscalls

* Error log: https://del.dog/ujofeqowol

### DIFF
--- a/tdutils/td/utils/port/detail/ThreadPthread.cpp
+++ b/tdutils/td/utils/port/detail/ThreadPthread.cpp
@@ -14,7 +14,9 @@ char disable_linker_warning_about_empty_file_thread_pthread_cpp TD_UNUSED;
 
 #include <pthread.h>
 #include <sched.h>
+#if TD_FREEBSD || TD_OPENBSD || TD_NETBSD
 #include <sys/sysctl.h>
+#endif
 #include <unistd.h>
 
 namespace td {


levlam:
@Akianonymus, thanks for pointing this out! We will definitely fix the issue, but your patch just removes include of the `sys/sysctl.h` header on all platforms. `SYS__sysctl` wouldn't be defined, unless `sys/syscall.h` is included first, but if the `sys/syscall.h` header is included first, then it will be defined even on Alpine Linux, because SYS__sysctl is a macros for Linux syscall and is defined there. So in that case the patch will not change anything.

Akianonymus:
@levlam Thx for pointing out, it was pretty dumb of me ! :(

I have removed the header inclusion completely, we don't need to include it here at all.

I have tested this commit in Ubuntu 14x, 16x and 18x and it builds fine.

Here is Ubuntu 18x build log: https://github.com/Akianonymus/td/commit/9186d79d0ae563350dbc01e5c9568576bae6a04e/checks?check_suite_id=248425063

levlam:
This is still wrong, because the header is needed on non-Linux systems. The correct fix is
```
#if TD_FREEBSD || TD_OPENBSD || TD_NETBSD
#include <sys/sysctl.h>
#endif
```

Akianonymus:
I have force pushed with your change and authorship.

isopen:
Levsky, Now, checks will be scattered throughout the code. Need a more general mechanism. Perhaps modules with new benefits will be just in time.

PR #757, opened 2019-10-28
fix wasm-utils

isopen:
It is unclear how this happens, but it works more efficiently.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ auto/
 db_backup
 *.pyc
 docs/
+.idea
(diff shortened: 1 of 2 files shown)


levlam:
@isopen Could you show exactly, what became faster after this change and how big is the speed up in different browsers?

isopen:
@levlam It was the default browser in Debian 10.
Mozilla 60

In all such cases, the mechanism is more efficient.

isopen:
@levlam https://developer.mozilla.org/en-US/docs/WebAssembly/Caching_modules
See what they are doing. Deprecated mechanisms without releasing new ones.
Still need to apply `WebAssembly.instantiateStreaming` for the caching mechanism. And the method via `instantiate` is done in the second try. Just in case.

PR #867, opened 2020-01-05
Added pkg config files for tdjson library

tanshihaj:


levlam:
@tanshihaj Have you tested it? Especially I'm interested in how linking to tdjson_static works without all dependent libs specified and how linking to tdjson works without stating that it depends on OpenSSL and zlib?

Thaodan:
A similar package config file is used for depecher on Sailfish OS and their it works.
Doesn't this already don't state it depends on openssl and zlib in non static mode?

levlam:
@Thaodan Dynamic libraries automatically load all dependent libraries, but static libraries can't work this way.

ff099261 (approved):

Marukesu:
I have written that .pc file
```
prefix=/usr/local
incluidedir=${prefix}/include
libdir=${prefix}/lib

Name: tdjson
Description: TdLib Json library
Version: 1.6.0

Requires.private: openssl, zlib
CFlags: -I${incluidedir}/td
Libs: -L${libdir} -ltdjson
```
if you use `pkg-config --cflags --libs --static tdjson` it actually passes the static dependencies:
```
$ pkg-config --cflags --libs --static tdjson
-I/usr/local/include/td -L/usr/local/lib -ltdjson -lssl -lcrypto -ldl -pthread -lz
```

levlam:
@Marukesu `tdjson` static library also depends on a lot of internal static libraries such as tdutils and tdcore. For example, see how it is statically linked to Go binary in https://github.com/zelenin/go-tdlib/blob/a5d1c7ffe705c2d89e1b9ec21c4d59e805540a64/client/tdjson_static.go#L8.

Marukesu:
so, a `tdjson-static.pc` will be
```
prefix=/usr/local
incluidedir=${prefix}/include
libdir=${prefix}/lib

Name: tdjson
Description: TdLib Json library (static)
Version: 1.6.0

Requires.private: zlib, openssl
CFlags: -I${incluidedir}/td
Libs: -L${libdir} -ltdjson_static -ltdjson_private -ltdclient -ltdcore -ltdactor -ltddb -ltdsqlite -ltdnet -ltdutils
Libs.private: -lstdc++ -lm
```
the pkg-config output:
```
$ pkg-config --cflags --libs --static tdjson-static 
-I/usr/local/include/td -L/usr/local/lib -ltdjson_static -ltdjson_private -ltdclient -ltdcore -ltdactor -ltddb -ltdsqlite -ltdnet -ltdutils -lstdc++ -lm -lz -lssl -lcrypto -ldl -pthread
```
but it's receiving various undefined reference from std

levlam:
All libraries should be private dependencies except `tdjson_static`.

Marukesu:
So i changed from libstdc++ to libc++ and libc++abi and it works for me here. One question, it will be possible to rename `libtdjson_static.a` to `libtdjson.a`? if so, it will be only one .pc file needed for both.

levlam:
We unlikely to change library name just for this.

levlam:
This was superseded by https://github.com/tdlib/td/pull/1223.

PR #877, opened 2020-01-09
Fix switch_pm_parameter base64 check

### LINKED ISSUE #876: Can tdlib handle InlineQuerys by Bot?
Is it with tdlib possible to use [InlineQuery ](https://core.telegram.org/bots/api#inlinequery)which is provided by a bot? In my case I made a Bot which provides a InlineQueryResultArticle which send a InlineButton to a user. So I would like to use this with tdlib...

Edit: I just found this: getInlineQueryResults function. Is this the way to achieve this?

### LINKED ISSUE #878: switch_pm_parameter base64 check is broken
Using Inline queries with a switch_pm_parameter is broken right now.
```Unallowed characters in switch_pm_parameter are used```

The bug popped up tonight at 3AM on all my bots. No changes on bot code or anything.
Inline queries seem to be broken for many bots right now. `switch_pm_parameter` doesn't work at all

@remembertheair found this change that has been made 5 days ago:
https://github.com/tdlib/td/commit/a9de33970b19f5dbc97f26bb907ced3e206b604e#diff-a1f80f20552f21f8b7ff9962e3772fadR333

Nukesor:
Resolves #878

### DIFF
--- a/td/telegram/InlineQueriesManager.cpp
+++ b/td/telegram/InlineQueriesManager.cpp
@@ -330,7 +330,7 @@ void InlineQueriesManager::answer_inline_query(int64 inline_query_id, bool is_pe
     if (switch_pm_parameter.size() > 64) {
       return promise.set_error(Status::Error(400, "Too long switch_pm_parameter specified"));
     }
-    if (!is_base64url(switch_pm_parameter)) {
+    if (!is_base64(switch_pm_parameter)) {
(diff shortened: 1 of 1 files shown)


levlam:
The proposed change is wrong, but thank you for trying. The original issue was fixed in https://github.com/tdlib/td/commit/d6dca91ff31494b097f876a31525f74b2f81c234.

PR #895, opened 2020-01-15
use CMAKE_INSTALL_LIBDIR variable instead of hardcoded lib in CmakeLi…

locez:
use ${CMAKE_INSTALL_LIBDIR} variable instead of hardcoded 'lib' in CmakeLists.txt
this is very useful for some system package manager( such as portage) to build the td library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,10 @@ cmake_minimum_required(VERSION 3.0.2 FATAL_ERROR)
 
 project(TDLib VERSION 1.5.4 LANGUAGES CXX C)
 
+if (NOT DEFINED CMAKE_INSTALL_LIBDIR)
+  set(CMAKE_INSTALL_LIBDIR "lib")
+endif(NOT DEFINED CMAKE_INSTALL_LIBDIR)
+
 if (POLICY CMP0054)
   # do not expand quoted arguments
   cmake_policy(SET CMP0054 NEW)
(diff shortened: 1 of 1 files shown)


levlam:
Thanks for pointing this out!

Is it much better to include https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html instead of the proposed manual setting of CMAKE_INSTALL_LIBDIR?

locez:
Actually, I'm not familiar with cmake, but I have tried to fix it.

levlam:
Thanks, @locez!

I figured out that despite including GNUInstallDirs gives much better default name for LIBDIR, it will break some things.

For example, in the [build instructions generator](https://tdlib.github.io/td/build.html?language=Java) there are two places where `lib/` directory is hardcoded, namely
https://github.com/tdlib/td/blob/92c9149e30318ac83baeb194ba9fcacf2bea3588/build.html#L824 and
https://github.com/tdlib/td/blob/92c9149e30318ac83baeb194ba9fcacf2bea3588/build.html#L848. It seems that there is no way to avoid this, so including GNUInstallDirs will break the instructions generator on some systems.

Also some TDLib-based clients like [telega.el](https://github.com/zevlg/telega.el) use hardcoded path `/usr/local/lib`, so they will be broken after the change.

Given all of this, let's stick with the first proposed commit, which only allows to manually specify CMAKE_INSTALL_LIBDIR.

locez:
I have test in my system, all above change runs right. telega.el also will search system's libary dir.

I agree you that only someone need this should set it, others will not change anythings

levlam:
Thanks, @locez!

I will merge that this week.

locez:
Thanks too! @levlam

PR #957, opened 2020-03-17
mention that getChats not works with Bots

jumoog:
getChats() does not work for bots and you have to use getChat() instead.

### DIFF
--- a/td/generate/scheme/td_api.tl
+++ b/td/generate/scheme/td_api.tl
@@ -3178,7 +3178,7 @@ getFile file_id:int32 = File;
 //@remote_file_id Remote identifier of the file to get @file_type File type, if known
 getRemoteFile remote_file_id:string file_type:FileType = File;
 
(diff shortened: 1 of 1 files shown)


levlam:
The TDLib's documentation is user-oriented. Most methods are not available for bots, but this is not reflected in the documentation on purpose to reduce its cluttering. When such a method is called, TDLib immediately returns the error "The method is not available for bots", so this should not be a significant problem for bot developers.

jumoog:
The problem is that the chat has to be known, so that you can send messages for example. From my point of view you should make an exception for this function because it is so important for other functions to work at all.

levlam:
@jumoog The usual rule is that through TDLib can be done exactly the same that through Bot API. In this particular case Bot API has method `getChat`, but has no ability to get all chats. Moreover, there are other ways to get a particular chat, for example, from an incoming from TDLib `updateNewChat` after receiving a message from the chat, or through `searchPublicChat`.

PR #962, opened 2020-03-18
Add missing Windows header (winioctl.h).

egorpugin:
`td-1.6.0/tdutils/td/utils/port/FileFd.cpp(213): error C2065: 'FSCTL_SET_SPARSE': undeclared identifier`

The latest MSVC (VS 2019 16.5) and WinSDK.
MSDN also says that FSCTL_SET_SPARSE is defined in `winioctl.h`.

### DIFF
--- a/tdutils/td/utils/port/FileFd.cpp
+++ b/tdutils/td/utils/port/FileFd.cpp
@@ -32,6 +32,10 @@
 #include <unistd.h>
 #endif
 
+#if TD_PORT_WINDOWS
+#include <winioctl.h>
+#endif
+
 namespace td {
 
 namespace {


levlam:
`winioctl.h` is an internal Windows SDK header. It should never be included directly and `Windows.h` needs to be included instead as documented in https://docs.microsoft.com/en-us/windows/win32/api/winioctl/ni-winioctl-fsctl_set_sparse. 

And `Windows.h` is globally included on Windows in `td/utils/common.h`. 

Does it compile successfully after your patch? What is your OS version?

egorpugin:
> Does it compile successfully after your patch?

Yes, and I'm using custom build.

> What is your OS version?

10.0.18363.720

> It should never be included directly and Windows.h

Oh, missed that. I'll try with Windows.h.

levlam:
For me `Windows.h` includes `winscard.h` unless someone defined `WIN32_LEAN_AND_MEAN` and `winscard.h` unconditionally includes `winioctl.h`.

What do you mean by "custom build"?

egorpugin:
Custom build system.

Ok, found that some dependency pushes WIN32_LEAN_AND_MEAN into td lib.
I'm wrong here, sorry.

levlam:
Thanks for the confirmation. I will consider supporting TDLib building with WIN32_LEAN_AND_MEAN macros defined, which is supposed to speed up build process on Windows.

PR #983, opened 2020-04-04
Added `tdlib` new php extension for tdlib

aurimasniekis:
Added `tdlib` new php extension for tdlib which integrates via tdjson JSON interface


levlam:
Thanks, @aurimasniekis!

PR #1092, opened 2020-06-12
Real-time build

isopen:
TDLib auto build.
https://github.com/isopen/td/runs/766384929?check_suite_focus=true
https://github.com/isopen/td/suites/791067731/artifacts/8441311

https://help.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions#about-billing-for-github-actions


isopen:
@levlam For public accounts is not always free.Rejected pull request

isopen:
@levlam It definitely needs gas from the paste to start 100

PR #1094, opened 2020-06-14
Using emcmake in build-tdlib.sh

isopen:


levlam:
Thanks. That looks legit.

isopen:
Yes. It seems like this moment is documented.
https://emscripten.org/docs/compiling/Building-Projects.html

PR #1124, opened 2020-07-05
improve chatInviteLinkInfo docs

mymike00:
The `chat_id` field in `chatInviteLinkInfo` claims to be 0 if the user isn't a member of the chat.
In fact, if the user isn't a member of the chat but the chat is public the `chat_id` is not 0 

If you find a better phrasing than mine, go ahead and edit it, I'm not that good with English...

### DIFF
--- a/td/generate/scheme/td_api.tl
+++ b/td/generate/scheme/td_api.tl
@@ -777,7 +777,7 @@ chatsNearby users_nearby:vector<chatNearby> supergroups_nearby:vector<chatNearby
 chatInviteLink invite_link:string = ChatInviteLink;
 
 //@description Contains information about a chat invite link
-//@chat_id Chat identifier of the invite link; 0 if the user is not a member of this chat
+//@chat_id Chat identifier of the invite link; 0 if the user is not a member of this chat and if the chat is not public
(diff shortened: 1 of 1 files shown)


levlam:
Thanks! You are right.

In fact, in the non-pushed to Github version the field description has already been fixed to "Chat identifier of the invite link; 0 if the user have no access to the chat before joining".

mymike00:
Ah, great to hear you already fixed it :+1: 

btw, why don't you push your changes right after you do them?

levlam:
@mymike00 TDLib source code is pushed once the corresponding version is suitable for general use. Current development version uses unstable server API, which breaks on a daily basis. API changes also often causes backward incompatible changes in TDLib during the development.

Currently, it is not possible to move the whole development to Github and even it is moved the corresponding version would be suitable for development only.

mymike00:
What about a `development` branch where you push the unstable code and then merge it into `master` when it gets more stable?

levlam:
@mymike00 As I said, this is currently not possible, because source code is extracted from the private repository using automated tool. Private and public git repositories has no common ancestor.

mymike00:
Ok, I understand. thank you for your help

PR #1212, opened 2020-09-27
Add callback-based logging stream

ColonelThirtyTwo:
Allows the user of the library to specify log functions, allowing the library to hook into other logging solutions.


levlam:
Thanks, @ColonelThirtyTwo!
This is a needed feature, a good way to add it and a good implementation. But I planned to add the feature in a different way though. TDLib already has the ability to hook fatal errors using `td_set_log_fatal_error_callback` and this new feature effectively extends that ability. I planned to allow log lines hooking in the same way, allowing the callback to receive all log messages. Then the same behavior would be achieved by setting the callback and setting the log stream to `logStreamEmpty`.

ColonelThirtyTwo:
Changing `td_set_log_fatal_error_callback` would break API+ABI compatibility. IDK how much that matters to you though.

levlam:
@ColonelThirtyTwo This is the reason, why the full replacement is planned only in TDLib 2.0, and for some time there should be two callback functions.

ColonelThirtyTwo:
If that is the case, are you willing to merge this?

levlam:
The feature will be added but in a different way, such as only one callback function is required for both log hooking and handling of fatal errors. The proposed changes will not be merged by that reason, but they are good by themselves and they can be used in the meanwhile.

ColonelThirtyTwo:
So what is the ETA for tdlib 2 then? If it's going to be any significant length of time, you may as well merge this so we can use it. It's not like its going to be an issue moving forward since V2 is not going to be backwards-compatible anyway.

levlam:
There is no ETA for future releases, but definitely 2.0 will not be released soon.

It's a good point that 2.0 will be anyway backward-incompatible, but if this pull request is merged, then there will be a point in time, when three callback functions, two of which are deprecated, exist. It is better to skip this state and have only one new callback function and one deprecated fatal error callback function.

r3z4r3557:
Hii

VChernomYA (commented):
[]()

levlam:
Finally, the function `td_set_log_message_callback` was added to TDLib 1.7.5, which allows to intercept all TDLib log entries. Combined with `logStreamEmpty` it allows to completely redirect TDLib's log anywhere. This addition supersedes the proposed changes.

PR #1219, opened 2020-10-01
Option "chat_unload_delay" for bots

Cavallium:
This option allows to override the default dialog unload delay.

The allowed range is from 0 to 86400 seconds (1 day).
Setting the value -1 restores the default behavior.
If the option is not set the default value will be -1, so it will use the default behavior.

### DIFF
--- a/td/telegram/MessagesManager.cpp
+++ b/td/telegram/MessagesManager.cpp
@@ -10303,11 +10303,11 @@ void MessagesManager::delete_all_channel_messages_from_user_on_server(ChannelId 
 int32 MessagesManager::get_unload_dialog_delay() const {
   constexpr int32 DIALOG_UNLOAD_DELAY = 60;        // seconds
   constexpr int32 DIALOG_UNLOAD_BOT_DELAY = 1800;  // seconds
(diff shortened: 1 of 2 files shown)


levlam:
The default behavior, should be `optionValueEmpty`, not a magic value.

In the TDLib API the term "chat" is used instead of "dialog".

Enabling dialog unload for non-bots without message database can lead to numerous issues and probably even some failed CHECKs.

Cavallium:
Ok, I'll fix

levlam:
set_integer_option has `min` and `max` parameters.

Cavallium:
Now the option is bounded from 60 seconds minimum to 1 day maximum.
Also, now you can set it only for bots and it will be read if is_bot is true.
The -1 value is no longer valid.

levlam:
Thanks!
Now the changes looks good. I will accept the pull request.

levlam:
Between, changes in `is_message_unload_enabled` aren't needed now.

Cavallium:
You are right, because before I used it to enable the unload for users.

levlam:
Thanks, @Cavallium!

I also renamed the option to "message_unload_delay", because chats itself are never unloaded, and simplified the implementation.

PR #1223, opened 2020-10-02
build pkg-config files for tdjson and tdjson_static

### LINKED ISSUE #1067: generate pkgconfig file
Some build tools (cabal for example) relies on pkg-config to discover c libraries. Would it be possible to generator this file when building tdlib?

Marukesu:
Fix #1067 
Supersede #867

### DIFF
--- a/CMake/GeneratePkgConfig.cmake
+++ b/CMake/GeneratePkgConfig.cmake
@@ -1,0 +1,57 @@
+function(get_relative_link REL PATH)
+	get_filename_component(NAME ${PATH} NAME_WE)
+	if(${NAME} MATCHES "^lib")
+		string(REGEX REPLACE "^lib" "-l" LINK ${NAME})
+	elseif(${NAME} MATCHES "^-")
+		set(LINK ${NAME})
+	else()
+		string(CONCAT LINK "-l" ${NAME})
+	endif()
+	set(${REL} ${LINK} PARENT_SCOPE)
+endfunction()
+
+function(generate_pkgconfig TARGET DESCRIPTION)
+	message("generating pkg-config for ${TARGET}")
+	get_filename_component(PREFIX ${CMAKE_INSTALL_PREFIX} ABSOLUTE)
+
+	get_target_property(LIST ${TARGET} LINK_LIBRARIES)
+	set(REQS "")
+	set(LIBS "")
+	foreach(LIB ${LIST})
+		if(TARGET ${LIB})
+			set(HAS_REQS 1)
+			list(APPEND REQS ${LIB})
+		else()
+			set(HAS_LIBS 1)
+			get_relative_link(LINK ${LIB})
+			list(APPEND LIBS ${LINK})
+		endif()
+	endforeach()
+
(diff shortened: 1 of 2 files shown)


levlam:
Thanks, @Marukesu!

This is substantially better than the previous pull request, but still there are a lot of drawbacks with this approach.

For example, if the compiler is clang, then libstdc++ still can be used as a standard C++ library. Also, it will be hard to manually mantain the "*.pc.in" files. The right approach would be to get the list of dependencies from CMake and convert it to pkg-config format.

Also, from the pkg-config point of view it seems reasonable to have a separate `.pc` file for all CMake targets, not only for `tdjson` and `tdjson_static`. And this is another reason, why they should be generated automatically using known to CMake data.

levlam:
The mentioned approach has been tried in https://stackoverflow.com/questions/44292462/how-to-auto-generate-pkgconfig-files-from-cmake-targets. The function implemented there needs a lot of improvements to generate a good pkg-config file, but the approach itself is right and much more flexible.

Marukesu:
Ok, now it's generating the files from the `generate_pkgconfig` function, it's linking to non-targets dependencies via the `Libs.private` field and Targets dependencies via `Requires.private` field.

for example, tdjson_static generates:
```
prefix=/usr/local
includedir=${prefix}/include
libdir=${prefix}/lib

Name: tdjson_static
Description: Telegram Library - Json Interface (static)
Version: 1.6.9

CFlags: -I${includedir}
Libs: -L${libdir} -ltdjson_static
Requires.private: tdjson_private
```
with pkg-config output:
```
$ pkg-config --cflags --libs --static tdjson_static
-I/usr/local/include -L/usr/local/lib -ltdjson_static -ltdjson_private -ltdclient -ltdcore -lcrypto -ldl -lz -ltdapi -ltdnet -ldl -lz -lssl -lcrypto -ltddb -ltdactor -ltdsqlite -lcrypto -ldl -lz -ltdutils -pthread -lcrypto -ldl -lz
```

since `libc++/libstdc++` are not a declared dependency it's not included in the files.

levlam:
Now I agree with the approach. I will test this and likely accept. Could you squash the commits and force push?

levlam:
Thanks, @Marukesu! Very good job.

levlam:
@Marukesu I identified and fixed some issues with the implementation. Now it seems to correctly handle most cases.

Now pkg-config output is:
```
$ pkg-config --cflags --libs --static tdjson_static
-I/usr/local/include -L/usr/local/lib -ltdjson_static -ltdjson_private -ltdclient -ltdcore -l:libcrypto.so -ldl -l:libz.so -ltdapi -ltdnet -l:libssl.so -l:libcrypto.so -ldl -l:libz.so -ltddb -ltdactor -ltdsqlite -l:libcrypto.so -ldl -l:libz.so -ltdutils -pthread -l:libcrypto.so -ldl -l:libz.so
```

The lack of `libc++/libstdc++` in the dependencies isn't a big issue, because if the code is built using `g++/clang++`, it will be added automatically.

PR #1227, opened 2020-10-04
Create README.md

dieletro:
I created a small example of using tdlib with Delphi and would like to make it available on the list

is this my github with exemple files https://github.com/dieletro/tdlib-delphi


levlam:
We do not add more directly new source code examples, because we will not able to maintain them. Only links to third-party repositiories in README can be accepted.

dieletro:
it would then be possible to create an Object Pascal / Delphi and another / Lazarus folder, each containing the readme with the necessary procedures for development

levlam:
No, we don't need more subdirectories for various languages, only a short description and links to the bindings in the README.md.

PR #1228, opened 2020-10-04
Create README.md

dieletro:
I created a small example of using tdlib with Lazarus and would like to make it available on the list

is this my github with exemple files https://github.com/dieletro/tdlib-lazarus

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -10,6 +10,8 @@ TDLib can be easily used from almost any programming language on any platform. S
 Choose your preferred programming language to see examples of usage and a detailed description:
 
 - [Python](#python)
+- [lazarus](#lazarus)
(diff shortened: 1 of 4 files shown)


levlam:
Not Lazarus, nor Delphi are programming languages. The section should have called "Object Pascal" and contain links to both Lazarus and Delphi projects if they are different. In the README there should be only links to the projects with the short description of them to keep reasonable the amount of information. Use entries for other languages as a reference.

dieletro:
Got it, I'll redo it this way

dieletro:
today Delphi is already considered a language, as it innovated and brought many new features to the language that Object Pascal itself does not have, but this is a matter of understanding, so I used it separately as a language

PR #1239, opened 2020-10-11
Delete CMakeLists.txt

sj5505:


msafari644:
Unfortunately, I am in Iran and the boycott region
But what does the Iranian government have to do with me ??
What is my fault ???

PR #1254, opened 2020-10-17
Minor bug fix in the cpp example

kolayne:
`std::cin` only reads the first word from the command-line, while Telegram passwords can contain spaces


levlam:
Thanks, @kolayne! This is a reasonable change.

Issue #1282, opened 2020-11-05
Error while compiling MessagesManager.cpp


jamel:
I'm trying to build telegram-bot-api with clang-8. But build failed on compiling MessagesManager.cpp:

```
~/src/telegram-bot-api/build [2020-11-05 23:46:42]
$ make
[  0%] Built target tdsqlite
[  0%] Built target generate_mime_types_gperf
[  1%] Built target tdmime_auto
[ 26%] Built target tdutils
[ 27%] Built target tdactor
[ 32%] Built target tddb
[ 32%] Built target memprof
[ 35%] Built target tdtl
[ 38%] Built target generate_common
[ 38%] Generate common tl source files
[ 38%] Built target tl_generate_common
[ 38%] Built target tdapi
[ 44%] Built target tdnet
[ 44%] Building CXX object td/CMakeFiles/tdcore.dir/td/telegram/MessagesManager.cpp.o
In file included from /home/jamel/src/telegram-bot-api/td/td/telegram/MessagesManager.cpp:7:
In file included from /home/jamel/src/telegram-bot-api/td/td/telegram/MessagesManager.h:11:
In file included from /home/jamel/src/telegram-bot-api/td/td/telegram/Dependencies.h:14:
In file included from /home/jamel/src/telegram-bot-api/td/td/telegram/WebPageId.h:11:
In file included from /home/jamel/src/telegram-bot-api/td/tdutils/td/utils/tl_helpers.h:21:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/unordered_set:47:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/bits/hashtable.h:35:
/usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/bits/hashtable_policy.h:85:11: error: implicit instantiation of undefined template 'std::hash<td::MessageContentType>'
        noexcept(declval<const _Hash&>()(declval<const _Key&>()))>
                 ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/type_traits:138:14: note: in instantiation of template class 'std::__detail::__is_noexcept_hash<td::MessageContentType, std::hash<td::MessageContentType> >' requested here
    : public conditional<_B1::value, _B2, _B1>::type
             ^
[...]

PR #1283, opened 2020-11-05
Implement hash function for MessageContentType enum.

jamel:
See issue #1282 for more details.


levlam:
Thanks! Could you use clang-format on the code?

jamel:
@levlam done

PR #1285, opened 2020-11-06
Fix code formatting.

isopen:
Minor fixes


levlam:
I don't like the proposed formatting style.

PR #1287, opened 2020-11-09
Add parameters specify tcp server bind ip address

KunoiSayami:
To specify [Bot Api server](https://github.com/tdlib/telegram-bot-api) bind address, must add parameters to `TcpListener` class in this library.

### DIFF
--- a/tdnet/td/net/TcpListener.cpp
+++ b/tdnet/td/net/TcpListener.cpp
@@ -11,7 +11,10 @@
 
 namespace td {
 
-TcpListener::TcpListener(int port, ActorShared<Callback> callback) : port_(port), callback_(std::move(callback)) {
+
+
+TcpListener::TcpListener(int port, ActorShared<Callback> callback, char const * server_address) :
+    port_(port), callback_(std::move(callback)), server_address_(server_address) {
 }
 
 void TcpListener::hangup() {
@@ -19,7 +22,7 @@ void TcpListener::hangup() {
 }
 
 void TcpListener::start_up() {
-  auto r_socket = ServerSocketFd::open(port_);
+  auto r_socket = ServerSocketFd::open(port_, CSlice(server_address_));
   if (r_socket.is_error()) {
     LOG(ERROR) << "Can't open server socket: " << r_socket.error();
     set_timeout_in(5);
--- a/tdnet/td/net/TcpListener.cpp
+++ b/tdnet/td/net/TcpListener.cpp
@@ -14,7 +14,7 @@ namespace td {
 
 
 TcpListener::TcpListener(int port, ActorShared<Callback> callback, Slice server_address) :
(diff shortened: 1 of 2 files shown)


tdnet/td/net/TcpListener.h
```diff
@@ -20,13 +20,14 @@ class TcpListener final : public Actor {
...
 
-  TcpListener(int port, ActorShared<Callback> callback);
+  TcpListener(int port, ActorShared<Callback> callback, char const * server_address = "0.0.0.0");
   void hangup() override;
 
  private:
   int port_;
   ServerSocketFd server_fd_;
   ActorShared<Callback> callback_;
+  char const * server_address_;
```
levlam, line 30: Storing a `const char *` field is very unsafe. The type of the field needs to be replaced with `string`.

tdnet/td/net/TcpListener.h
```diff
@@ -20,13 +20,14 @@ class TcpListener final : public Actor {
     virtual void accept(SocketFd fd) = 0;
   };
 
-  TcpListener(int port, ActorShared<Callback> callback);
+  TcpListener(int port, ActorShared<Callback> callback, char const * server_address = "0.0.0.0");
```
levlam, line 23: Use `Slice` instead of `const char *` as the parameter type as a more general type: `Slice server_address = Slice("0.0.0.0")`.

levlam (changes requested):

tdnet/td/net/TcpListener.cpp
```diff
@@ -13,8 +13,8 @@ namespace td {
 
 
 
-TcpListener::TcpListener(int port, ActorShared<Callback> callback, char const * server_address) :
-    port_(port), callback_(std::move(callback)), server_address_(server_address) {
+TcpListener::TcpListener(int port, ActorShared<Callback> callback, Slice server_address) :
+    port_(port), callback_(std::move(callback)), server_address_(server_address.data()) {
```
levlam, line 17: `Slice` isn't null-terminated, so It is wrong to use `.data()` to convert if to `string`. Use `.str()` instead.

tdnet/td/net/TcpListener.cpp
```diff
@@ -11,15 +11,18 @@
...
+    port_(port), callback_(std::move(callback)), server_address_(server_address) {
 }
 
 void TcpListener::hangup() {
   stop();
 }
 
 void TcpListener::start_up() {
-  auto r_socket = ServerSocketFd::open(port_);
+  auto r_socket = ServerSocketFd::open(port_, CSlice(server_address_));
```
levlam, line 25: Explicit conversion to `CSlice` isn't needed anymore.

levlam (changes requested):

levlam (approved):

PR #1301, opened 2020-11-19
Update build.html

darmiel:
Due to a license change, 'java' isn't available in brew cask anymore. 
But you can still install java using AdoptJDK java.
```bash
brew tap adoptopenjdk/openjdk
brew cask install adoptopenjdk
```

The Brew installation script is now Bash and no longer Ruby.

### DIFF
--- a/build.html
+++ b/build.html
@@ -492,7 +492,7 @@ function onOptionsChanged() {
     if (target !== 'C++/CLI' && target !== 'C++/CX') {
       pre_text.push('Note that Windows Subsystem for Linux (WSL) and Cygwin are not Windows environments, so you need to use instructions for Linux for them instead.');
     }
-    pre_text.push('Download and install <a href="https://visualstudio.microsoft.com/ru/vs/community/">Microsoft Visual Studio</a>. Enable C++' + win10_sdk + ' support while installing.');
+    pre_text.push('Download and install <a href="https://visualstudio.microsoft.com/vs/community/">Microsoft Visual Studio</a>. Enable C++' + win10_sdk + ' support while installing.');
     pre_text.push('Download and install <a href="https://cmake.org/download/">CMake</a>; choose "Add CMake to the system PATH" option while installing.');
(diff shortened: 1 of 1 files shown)


levlam:
Thanks!

I see that `java` was migrated to "homebrew/core" in https://github.com/Homebrew/homebrew-cask/commit/4e23f9fda882d2b183dce6f037a093aead454f3a and not removed, so we can use just `brew install openjdk`. Can we?

darmiel:
Oh, you're right. I hadn't noticed that!

levlam:
@darmiel Does `brew install java` works for you? I can find only `openjdk` package in brew. It looks like the moved package changed its name.

darmiel:
I tested it locally and with GitHub Actions: https://github.com/darmiel/macos-brew-java-test/runs/1432950346
However, the macos-latest image of GitHub uses some taps for Brew. Local I have:
```
homebrew/cask
homebrew/cask-versions
homebrew/core
homebrew/services
```

I think it would still be a better idea to use "openjdk" in case I (and GitHub) use some tap that makes java an alias of openjdk.

levlam:
Thanks, @darmiel!

levlam:
Between, I found package "java" in homebrew/core, which is an alias to "openjdk", so both package names are correct: https://github.com/Homebrew/homebrew-core/blob/master/Aliases/java.

darmiel:
Hey @levlam 

Sorry to misuse the pull request for another problem, but I believe that the problem is not the TdLib.

I would like to receive the info from a Supergroup / Basicgroup. 
For this I use GetSupergroup, or GetBasicGroup.

But they expect a 32-bit integer as parameter, the chat id from a message however,  of is a 64-bit integer:
```java
  @Subscribe
  public void onNewMessage(final UpdateNewMessage update) {        
    final long chatId = update.message.chatId;
    this.client.send(
        new GetSupergroup(/* <- 32 Bit */ chatId /* <- 64 Bit */),
        new ChatInfoResultHandler()
    );
  }
```

What do I have to use to get the ID of the supergroup?

PR #1302, opened 2020-11-20
Increase version to 1.7.0.

jumoog:
after updating change log also update version


levlam:
Updating Change Log is a small part of the release. The version will be updated when the release is ready.

PR #1336, opened 2020-12-21
Remove extra include dir part.

egorpugin:
Looks like an artifact of copy paste or something like that.
It forces us to add additional include directory to the root of tdlib sources `-I.` near with correct `-Itddb`.
Grep shows that this is the single source file with `/td/` (or `\/td\/`) pattern that adds something before root `td/` part.

### DIFF
--- a/td/telegram/logevent/LogEventHelper.cpp
+++ b/td/telegram/logevent/LogEventHelper.cpp
@@ -9,7 +9,7 @@
 #include "td/telegram/Global.h"
 #include "td/telegram/TdDb.h"
 
-#include "tddb/td/db/binlog/BinlogHelper.h"
+#include "td/db/binlog/BinlogHelper.h"
 
 #include "td/utils/logging.h"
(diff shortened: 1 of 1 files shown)


levlam:
This is a legitimate change, but it can't affect in any way compiler options. There is no way source code can affect external compiler options.
`-I.` is required by all files in `td/...`, because they include other files in that folder.

egorpugin:
> but it can't affect in any way compiler options

I mean, what is include dir root? With what part does every include start? I assume with `td/` and not `tddb/td` or `tdactor/td` etc.
I had build script for 1.6.0, everything was built fine.
On 1.7.0 this misused include appeared and the build is broken. This is the single build error.

> -I. is required by all files in td/..., because they include other files in that folder.

I build everything in separate packages. Every target is moved to its own folder, so no common `-I.` include is set.
You can check build specs (in C++) here https://software-network.org/org.sw.demo.telegram.td.core/specification

PR #1351, opened 2020-12-31
Added .NET reference to https://github.com/egramtel/tdsharp

JosXa:


levlam:
There are 2 mentions of `egramtel` in https://github.com/tdlib/td/tree/master/example#using-tdlib-in-c-projects. Main README is supposed to contain only documentation for native interfaces provided by TDLib itself.

JosXa:
Wasn't aware, and makese sense :)

PR #1464, opened 2021-03-25
Fix memory leak of "handlers_" in cpp example

### LINKED ISSUE #1461: When should I erase the key-value in the "handlers" map?
https://github.com/tdlib/td/blob/c6dec21cb691f1a47b2f97f1c2fb4217af6a154d/example/cpp/td_example.cpp#L156-L176

I found that after calling `send_query` with a `handler`, memory will always increase and not decrease, which seems to be directly related to `handlers_`, the example always inserts the key-value into `handlers_` without erasing it.

Is this a bug? When should I erase key-values that are no longer needed? Does `tdlib` guarantee to always reply to me only one `response` after I call `send`? If so, I can erase the key-value after calling the `handler` once.

If this is not a bug, how should I mitigate this problem?

Thanks in advance!

SpriteOvO:
Fix #1461

### DIFF
--- a/example/cpp/td_example.cpp
+++ b/example/cpp/td_example.cpp
@@ -172,6 +172,7 @@ class TdExample {
     auto it = handlers_.find(response.request_id);
     if (it != handlers_.end()) {
       it->second(std::move(response.object));
+      handlers_.erase(it);
     }
   }
 


levlam:
Thanks, @SpriteOvO!

PR #1467, opened 2021-03-27
Use RelWithDebInfo for UWP package

FrayxRulez:
**Why**
Currently the UWP package is built for both `Debug` and `Release` configurations.
This doesn't only double build times, but it also makes really hard to find production issues.

**What**
This PR aims to resolve the problem by building UWP package using `RelWithDebInfo` configuration, that is in fact a `Release` with PDB symbols.


levlam:
RelWithDebInfo can be used on Windows instead of Release, but Debug configuration needs to be kept. Debug configuration is much more suitable for debugging than any Release configuration.

FrayxRulez:
I added `-release_only` flag to build the package using `RelWithDebInfo` configuration.
Omitting the flag with build the usual Debug + Release package.

VChernomYA (approved):

PR #1542, opened 2021-05-23
Make BUILD_SHARED_LIBS yield dynamic lib that doesn't need static libs

pontaoski:
By changing the mode of a few libraries from STATIC to OBJECT
when BUILD_SHARED_LIBS is enabled, CMake will output a shared
library object that doesn't require any static linking against
Td from the library user's end.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -683,7 +683,7 @@ if (MEMPROF)
 endif()
 
 
-add_library(tdapi STATIC ${TL_TD_API_SOURCE})
+add_library(tdapi ${TL_TD_API_SOURCE})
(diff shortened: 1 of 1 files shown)


levlam:
Object libraries can be linked to with target_link_libraries() only since CMake 3.12 and TDLib supports CMake 3.0+, so they can't be used this way.

pontaoski:
> Object libraries can be linked to with target_link_libraries() only since CMake 3.12 and TDLib allows CMake 3.0+, so they can't be used this way.

I guess I can add a version shield so that OBJECT is only used when CMake version >= 3.12

levlam:
Thanks. Now this is a correct change. It will be merged in TDLib 1.7.5.

PR #1600, opened 2021-07-08
Clarify importance of generated files on non-macOS builds

Kylmakalle:
I've tried to build `TDLib`for iOS and found out that some files are generated only for macOS builds. 

I want to propose a change for docs because i've naturally skipped pregeneration docs part due to weak "`can`" word. 

I've tried to build `TDLib` for iOS only and failed because some **required** files were missing.
So we definitely **`must`** pregenerate files if we don't want to build macOS lib in a same flow (pipeline, etc)

Example behaviour below

### ❌ `./build.sh` platforms="iOS"
error, unable to find some generated tdutils file
image

### ✅ `./build.sh` platforms="macOS iOS watchOS"
sources generated "on the fly" in same build phase
image

### ✅ prepare_cross_compiling
sources generated with separate command
image

### DIFF
--- a/example/ios/README.md
+++ b/example/ios/README.md
@@ -13,7 +13,8 @@ To compile `TDLib` you will need to:
 ```
 brew install gperf cmake coreutils
 ```
-* If you don't want to build `TDLib` for macOS, you **must** pregenerate required source code files using `CMake` prepare_cross_compiling target:
+* `TDLib` requires some generated source code files. Those files generates automatically if you're building `TDLib` for macOS.
+If you don't want to build `TDLib` for macOS _or_ building it separatelly _or_ building it not in first place _(e.g. in order: iOS, watchOS, macOS)_, you **must** pregenerate required source code files using `CMake` prepare_cross_compiling target:
 ```
 cd <path to TDLib sources>
 mkdir native-build
--- a/example/ios/README.md
+++ b/example/ios/README.md
@@ -13,7 +13,7 @@ To compile `TDLib` you will need to:
 ```
 brew install gperf cmake coreutils
 ```
-* If you don't want to build `TDLib` for macOS, you can pregenerate required source code files using `CMake` prepare_cross_compiling target:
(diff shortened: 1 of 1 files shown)


levlam:
I agree that despite there are a lot of ways to generate needed source files, "must" is more appropriate there than "can".

But the readme must be as short as possible to be actually read. Given that the paragraph is optional and isn't needed at all for most users, I would stick with the changes from the first commit. Instead of the second commit, I propose to write just "to build `TDLib` for macOS first" instead of "to build `TDLib` for macOS".

Kylmakalle:
Thank you. I was not sure about big paragraph either.
I will leave some keywords below so inclined users can find this thread later

Kylmakalle:
Error that could be faced if no sources were generated. Make sure to use `prepare_cross_compiling` or compile `macOS` platform first!

```
make[2]: *** No rule to make target `../../../../tdutils/generate/auto/mime_type_to_extension.cpp', needed by `tdutils/CMakeFiles/tdutils.dir/generate/auto/mime_type_to_extension.cpp.o'.  Stop.
make[2]: *** Waiting for unfinished jobs....
[ 14%] Building CXX object tdutils/CMakeFiles/tdutils.dir/td/utils/port/detail/WineventPoll.cpp.o
make[1]: *** [tdutils/CMakeFiles/tdutils.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
td/sqlite/sqlite/sqlite3.c:38342:8: warning: "gethostuuid() is disabled." [-W#warnings]
#      warning "gethostuuid() is disabled."
       ^
1 warning generated.
td/sqlite/sqlite/sqlite3.c:38342:8: warning: "gethostuuid() is disabled." [-W#warnings]
#      warning "gethostuuid() is disabled."
       ^
1 warning generated.
[ 14%] Linking C static library libtdsqlite.a
[ 14%] Built target tdsqlite
make: *** [all] Error 2
```

Kylmakalle:
@levlam done 👌

levlam:
Thanks. I will be able to merge this before the next version increase.

PR #1638, opened 2021-07-29
Compilation issue

Cavallium:
Fix the following compilation issue:

td/telegram/ConfigManager.cpp:1693:25: error: ordered comparison between pointer and zero ('telegram_api::JSONValue *' and 'int')
              if (value > 0) {
                  ~~~~~ ^ ~


levlam:
Thanks. This is a right fix, but I already did the same change before spotting the pull request.

PR #1665, opened 2021-09-03
Remove unused inclusions

d351d3r:
Client.h contain 
```
#include "td/telegram/td_api.h"
#include "td/telegram/td_api.hpp"
```
And `std::vector `not used


levlam:
td_api.h and td_api.hpp are extensively used throughout the example code.

std::vector is implicitly used. For exmaple, chats->chat_ids_ is a vector.

PR #1689, opened 2021-09-28
Improve appearance of build instructions generator

Mammad900:
Improved formatting, colors and other appearance-related things.  
Also added a dark mode which is activated based on the user's system preferences.

Before:
image

After: 
image
image


Muaath5 (approved):
Almost all of commits are CSS editing.
The only thing was edited in Javascript is using `<ul>` instead of `\n`
And `<code>` instead of `<pre>`.

It supports dark mode, And uses CSS Variables.

Pearlminima:
> Almost all of commits are CSS editing.
> 
> The only thing was edited in Javascript is using `<ul>` instead of `\n`
> 
> And `<code>` instead of `<pre>`.
> 
> 
> 
> It supports dark mode, And uses CSS Variables.

cavallium:
`<meta name="viewport" content="width=device-width, initial-scale=1">` is missing

cavallium:
Also `<meta charset="UTF-8">` and `lang=en` are missing

PR #1694, opened 2021-10-03
CMakeLists: show error in case libraries not found

ForNeVeR:
This will cause CMake to exit with an error exit code instead of returning zero.

I've found it returns zero after a recent change on GitHub Actions where they've broken OpenSSL detection by the default TDLib build scripts: see https://github.com/actions/virtual-environments/issues/4195.

Without this change, my GitHub Actions build script which calls `cmake && cmake --build` ran successfully but just not generated any binaries on macOS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,7 +173,7 @@ if (NOT CMAKE_CROSSCOMPILING)
 endif()
 
 if (NOT OPENSSL_FOUND)
-  message(WARNING "Can't find OpenSSL: stop TDLib building")
+  message(FATAL_ERROR "Can't find OpenSSL: stop TDLib building")
   return()
 endif()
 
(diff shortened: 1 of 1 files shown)


levlam:
The warning instead of a fatal error is intended, because it should be possible to build `prepare_cross_compiling` target without having OpenSSL and zlib installed.

ForNeVeR:
I see. How could we improve that, then?

Perhaps the `install` target shouldn't finish successfully if the configuration hasn't finished properly?

FulgurIgor:
@levlam, why do you not check the usage cross-compilers when you want to show warnings instead of errors?

levlam:
@ForNeVeR There seems to be no way to disable install target in CMake.

levlam:
@FulgurIgor Because `prepare_cross_compiling` target is never used and can't be used during cross compliation. It must be used with a native compiler before cross compilation starts.

ForNeVeR:
Ok, sorry, I don't know a way to fix the issue without breaking the cross-compilation workflow.

levlam:
@ForNeVeR You likely need to build a specific target, for example, tdjson. Then you can use
```
cmake --build . --target tdjson || exit 1
cmake --build . --target install || exit 1
```

PR #1728, opened 2021-11-10
Add missing 'struct' keywords for tdc headers

Trumeet:
Some structs do not have typedef aliases, but their references forget to put 'struct' keywords when referencing them, so they cause compiling issues.

This commit adds the missing 'struct' keywords so TDLib C headers could compile successfully.


levlam:
Thanks, @Trumeet!

PR #1743, opened 2021-11-18
Update python example: Use sys.exit for quitting

sebix:
automatically sends the message to `stderr` and sets the exit-code to 1
`quit` is only for exiting the interpreter, use the recommended `sys.exit`


levlam:
You have a bug in your changes. You can't pass two arguments to sys.exit:
```
sys.exit('TDLib fatal error: ', message)
```

sebix:
> You have a bug in your changes. You can't pass two arguments to sys.exit:

Of course, thanks. Replacing `print` by `sys.exit` doesn't work.

-> Fixed

levlam:
Thanks, @sebix!

PR #1780, opened 2021-12-07
MacOS OpenSSL path fix

Sasha2018RL:
Fixed MacOS OpenSSL path. If install openssl using brew - it installed to /opt/homebrew/opt/openssl@3/

### DIFF
--- a/build.html
+++ b/build.html
@@ -957,7 +957,7 @@ function onOptionsChanged() {
 
     cmake_init_options = getBacicCmakeInitOptions();
     if (os_mac) {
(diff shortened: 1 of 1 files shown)


levlam:
`/opt/homebrew` path is specific for Apple Silicon. We seems to need separate build instructions for Apple Silicon laptops.

Also, isn't there a symlink from "/opt/homebrew/opt/openssl/" to "/opt/homebrew/opt/openssl@3/"?

Sasha2018RL:
> /opt/homebrew/opt/openssl/

image

levlam:
I added the ability to choose between Apple silicon and Intel when generating build instructions. Now the instructions should be correct for both architectures.

PR #1852, opened 2022-02-07
fix broken url

ivk1800:
extra dot ad the end broke the link. command + right click on the link in IDE open broken web page.


levlam:
There must be a dot at the end of the sentence. The URL isn't broken at the official documentation: https://core.telegram.org/tdlib/docs/classtd_1_1td__api_1_1set_option.html, so it is rather URL highlighting in IDE is broken.

PR #1884, opened 2022-03-13
Add more Settings links

alyral:
This pull request adds support for more `tg://settings` links:
* `tg://settings/theme` - from the iOS client
* `tg://settings/language` - from Telegram Desktop
* `tg://settings/privacy` - from the native macOS client

These could be defined "non-standard" links, but some of the already implemented ones (such as `tg://settings/change_number`) are non-standard as well.


levlam:
Other two links also weren't supported intentionally, but I don't hesistate their addition.

If you change the following, the commit will be merged.
1. `internalLinkTypePrivacySettings` needs to be renamed to `internalLinkTypePrivacyAndSecuritySettings`.
2. New types and functions needs to be moved to the correct place in lexicographical order in td_api.tl, LinkManager.cpp, LinkManager.h and test/link.cpp.

alyral:
> If you change the following, the commit will be merged.

Done, I applied those fixes. Would it be possible to also add the `tg://scanqr` and `tg://calllog` links (from Android), before merging?

PR #1950, opened 2022-04-24
Add session type field

alyral:
This pull request adds a new `type` field to the `session` class, which can be used by graphical clients to display a corresponding icon.

*Currently, apps are supposed to parse session parameters themselves to detect its type, which can lead to inconsistencies.*

I used the same conditions as the official [iOS](https://github.com/TelegramMessenger/Telegram-iOS/blob/f95cfbdc42bee2b4ee92750ca2bda0e85d68e942/submodules/SettingsUI/Sources/Privacy%20and%20Security/Recent%20Sessions/ItemListRecentSessionItem.swift#L113) and [macOS](https://github.com/overtake/TelegramSwift/blob/13d62c1c03df7dea70bad0f395fb68115f5a0b43/Telegram-Mac/RecentSessionsController.swift#L19) apps, and [Unigram](https://github.com/UnigramDev/Unigram/blob/4870c5ef42d757c93297da2fbafa29fc04c97f38/Unigram/Unigram/Controls/Cells/SessionCell.xaml.cs#L42), with some improvements.

### DIFF
--- a/td/generate/scheme/td_api.tl
+++ b/td/generate/scheme/td_api.tl
@@ -3281,18 +3281,73 @@ userPrivacySettingAllowFindingByPhoneNumber = UserPrivacySetting;
 accountTtl days:int32 = AccountTtl;
 
 
+//@class SessionType @description Represents the type of a session
+
+//@description This session is running on an Android device
+sessionTypeAndroid = SessionType;
+
+//@description This session is running on a generic Apple device
+sessionTypeApple = SessionType;
+
+//@description This session is running on the Brave browser
+sessionTypeBrave = SessionType;
+
+//@description This session is running on the Chrome browser
+sessionTypeChrome = SessionType;
+
+//@description This session is running on the Edge browser
+sessionTypeEdge = SessionType;
+
+//@description This session is running on the Firefox browser
+sessionTypeFirefox = SessionType;
+
+//@description This session is running on an iPad device
+sessionTypeIpad = SessionType;
+
+//@description This session is running on an iPhone device
+sessionTypeIphone = SessionType;
+
+//@description This session is running on a Linux device
+sessionTypeLinux = SessionType;
+
+//@description This session is running on a Mac device
+sessionTypeMac = SessionType;
+
+//@description This session is running on the Opera browser
+sessionTypeOpera = SessionType;
+
+//@description This session is running on the Safari browser
+sessionTypeSafari = SessionType;
+
+//@description This session is running on an Ubuntu device
+sessionTypeUbuntu = SessionType;
+
+//@description This session is running on an unknown type of device
+sessionTypeUnknown = SessionType;
+
+//@description This session is running on the Vivaldi browser
+sessionTypeVivaldi = SessionType;
+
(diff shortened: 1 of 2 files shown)


levlam:
Thank you.

This definitely needs to be added and the implementation is good. I have a few comments.

The code
> get_session_type(std::move(authorization)), authorization->api_id_,

is valid, but it can be very dengerous. Could you pass `authorization` by a const reference to `get_session_type` instead?

Also, could you add inside `get_session_type` lambda
```
auto contains = [](const string &str, const char *substr) {
  return str.find(substr) != string::npos;
};
```
and use it whenever appropriate?

App name isn't suitable for platform determination. It can be used as a hint, but it shouldn't be used as the primary factor.

alyral:
> App name isn't suitable for platform determination. It can be used as a hint, but it shouldn't be used as the primary factor.

The problem is that no other session data seems to indicate whether it's running on a web client, and some browser names might cause conflicts (e.g. Edge and some Samsung devices). Do you have any alternative about this?

levlam:
> Do you have any alternative about this?

Lets do a case sensitive check for "Web" + checking that the next letter isn't a small English letter to avoid matching "web" as substring.

I would also use `begins_with` instead of `contains` and case sensitive search whenever appropriate. It is better to return `generic` instead of a wrong one.

Also, why is it "edg" and not "edge"?

CrisMystik:
Thanks for your reply, I will implement these changes as soon as possible.

> Also, why is it "edg" and not "edge"?

Edge's identifier in the User Agent is `Edg` on Desktop, `EdgA` on Android, `EdgiOS` on iOS, and `Edge` for the legacy version.

Usually User Agent strings are parsed on the server-side, but not always (in my experience), so it seemed reasonable to just use `edg` to me.

Unfortunately I can't try the behaviour of the Android app (which uses `edge`), because for some reason the server detects Edge on Android as Chrome.

alyral:
Thanks for your reply, I will implement these changes as soon as possible.

>Also, why is it "edg" and not "edge"?

Edge's identifier in the User Agent is `Edg` on Desktop, `EdgA` on Android, `EdgiOS` on iOS, and `Edge` for the legacy version.

Usually User Agent strings are parsed on the server-side, but not always (in my experience), so it seemed reasonable to just use edg to me.

Unfortunately I can't try the behaviour of the Android app (which uses edge), because for some reason the server detects Edge on Android as Chrome.

*Sorry if I have sent you the same notification for two times, I've used the wrong GitHub account before.*

levlam:
> Where, exactly?

For example, I expect `platform` to begin with "Android"/"Windows"/"Ubuntu" to actually be one of them.

Could you also replace `web` initialization with lambda for clarity and use `const string &` type for `app_name` to avoid copying:
```
  bool web = [&] {
    Slice web_name("Web");
    auto pos = app_name.find(web_name.c_str());
    if (pos == string::npos) {
      return false;
    }

    auto next_character = app_name[pos + web_name.size()];
    return !('a' <= next_character && next_character <= 'z');
  }();
```

levlam:
Thank you, @alyral!

PR #1980, opened 2022-05-21
Updates README.

innerNULL:
Appends another cpp integration approach with cmake.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,20 @@ target_link_libraries(YourTarget PRIVATE Td::TdStatic)
 ```
 See [example/cpp/CMakeLists.txt](https://github.com/tdlib/td/tree/master/example/cpp/CMakeLists.txt).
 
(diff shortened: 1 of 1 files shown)


levlam:
The proposed changes are CMake-specific and aren't unique for TDLib, It would be too verbose to enumerate all ways to include and depend on a project in CMake.

Moreover, I would shrink the section "Using in CMake C++ projects" even more, because most TDLib users doesn't use it from C++, and the others are very likely to know, how to use CMake.

innerNULL:
> The proposed changes are CMake-specific and aren't unique for TDLib, It would be too verbose to enumerate all ways to include and depend on a project in CMake.
> 
> Moreover, I would shrink the section "Using in CMake C++ projects" even more, because most TDLib users doesn't use it from C++, and the others are very likely to know, how to use CMake.

sure~thx

PR #2015, opened 2022-06-18
update example/readme.md

azkadev:
added sample library and sample application telegram in flutter and dart


levlam:
I can't see any actual TDLib usage in these projects.

PR #2210, opened 2022-11-25
Fix MinGW threading compilation error

EricKotato:
This PR fixes two compilation error in MinGW-w64 under MSYS:

1. **Thread compilation error: invalid conversion**
   Same problem has been found here: https://github.com/boostorg/fiber/pull/281. In TDLib it happens in `tdutils/td/utils/port/detail/ThreadStl.h` and adding `reinterpret_cast<HANDLE>` fixes it.
2. **cc1plus.exe: out of memory**
   This happens only in 32-bit MinGW build. `td/telegram/MessagesManager.cpp` gives an out of memory error. There is an [old compiler bug in GCC](https://sourceforge.net/p/mingw-w64/mailman/message/33182613/) which either wasn't fixed or reappeared again. Adding `-ftrack-macro-expansion=0` to compiler flags solves the problem.
   Moved to #2212 by request.

PR #2212, opened 2022-11-25
Fix MinGW (32-bit) out of memory error

EricKotato:
Moved from #2210 with changes suggested in its comments.

This PR fixes `cc1plus.exe: out of memory` error in 32-bit MinGW build. td/telegram/MessagesManager.cpp gives an out of memory error. There is an [old compiler bug in GCC](https://sourceforge.net/p/mingw-w64/mailman/message/33182613/) which either wasn't fixed or reappeared again. Adding `-ftrack-macro-expansion=0` to compiler flags solves the problem.

### DIFF
--- a/CMake/TdSetUpCompiler.cmake
+++ b/CMake/TdSetUpCompiler.cmake
@@ -155,6 +155,10 @@ function(td_set_up_compiler)
     add_cxx_compiler_flag("-Wno-return-stack-address")
   endif()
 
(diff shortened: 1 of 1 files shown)


levlam:
I see absolutely no impact of the option on RAM consumption for GCC 6.3, but as soon as it helped you, it is safe to add it for MinGW.

PR #2264, opened 2023-01-10
Tcp transport 'Abridget' size check

landgrafhomyak:


levlam:
The proposd change makes no sense.

levlam:
`AbridgedTransport` has never been used in TDLib, so it is unlikely to work.

landgrafhomyak:
Why there then is unused code in repo?

levlam:
Unreachable code doesn't harm.

PR #2288, opened 2023-01-30
Add update version script.

AYMENJD:
I've made a bash script that update's version number a cross some files if the current TDLib version in `CMakeLists.txt` is modified. And then `git commit` it with the correct commit message `Update version to X.Y.Z.`.

By default, the script updates:
- [`README.md`](https://github.com/tdlib/td/blob/master/README.md).
- [`example/cpp/CMakeLists.txt`](https://github.com/tdlib/td/blob/master/example/cpp/CMakeLists.txt).
- [`example/uwp/extension.vsixmanifest`](https://github.com/tdlib/td/blob/master/example/uwp/extension.vsixmanifest).
- [`td/telegram/OptionManager.cpp`](https://github.com/tdlib/td/blob/master/td/telegram/OptionManager.cpp).

And you can easily drop these files from changing version by specifying argument `-i`. You can also pass unlimited number of files to change.

After making changes to these files, the script will execute `git diff [FILES...]` which will help you to ensure that everything is changed correctly. 

It will ask you `(y/n)` question if you want to commit these changes:
- If `Y/y` everything is okay, it will `git commit` the files.
- If `N/n` there is something wrong, undo changes and exit.

All thanks goes to @levlam.


levlam:
Thank you, @AYMENJD!

PR #2354, opened 2023-03-16
Fix a type on viewMessages method documentation.

john-preston:
The `source` argument should be marked as maybe null, not the `force_reply` argument.


levlam:
Thank you!

PR #2362, opened 2023-03-22
Added nupkg support to UWP example

FrayxRulez:
Should we remove `-release_only` in favor of `-nupkg`?


levlam:
> Should we remove -release_only in favor of -nupkg?

No. It is good to have release-only VSIX.

levlam:
Thank you!

Issue #2499, opened 2023-06-15
IPv6 auto-switching


vincentneo:
According to #63:
>Currently, no, because the **library should automatically switch to IPv6 connection if it is better than IPv4 one.** But we may add such option in a future, because it can slightly speed up connection creation time in case of unreliable IPv4.
            
Is that still the expected behaviour for TDLib? If so, I think I have another bug report to file, likely targeted for watchOS/HTTP connection only.

I realised that `prefer_ipv6` must be explicitly set to `true`, in order to operate via IPv6.

My test was very simple (correct me if this way of testing is flawed):
1. Set Configure IPv4 to `Off` on Mac settings, leaving only the Configure IPv6 with `Automatically`.
2. Test that IPv6 connections are still possible by browsing http://ipv6.test-ipv6.com/index.html.en_US
3. Run a watchOS app on simulator with TDLib (latest master) and observe:

```
2023-06-15 23:51:45.945774+0800 Pigeon[82091:901856] PDTask <6B294476-7971-46B9-9274-719625D4BAA0>.<14> finished with error [-1009] Error Domain=NSURLErrorDomain Code=-1009 "The Internet connection appears to be offline." UserInfo={_kCFStreamErrorCodeKey=50, _NSURLErrorFailingURLSessionTaskErrorKey=LocalDataPDTask <6B294476-7971-46B9-9274-719625D4BAA0>.<14>, _NSURLErrorRelatedURLSessionTaskErrorKey=(
    "LocalDataPDTask <6B294476-7971-46B9-9274-719625D4BAA0>.<14>",
    "LocalDataTask <6B294476-7971-46B9-9274-719625D4BAA0>.<14>"
), NSLocalizedDescription=The Internet connection appears to be offline., _kCFStreamErrorDomainKey=1, NSErrorFailingURLStringKey=http://149.154.171.5:443/api, NSErrorFailingURLKey=http://149.154.171.5:443/api}
2023-06-15 23:51:45.946227+0800 Pigeon[82091:901856] Task <6B294476-7971-46B9-9274-719625D4BAA0>.<14> finished with error [-1009] Error Domain=NSURLErrorDomain Code=-1009 "The Internet connection appears to be offline." UserInfo={_kCFStreamErrorCodeKey=50, _NSURLErrorFailingURLSessionTaskErrorKey=LocalDataTask <6B294476-7971-46B9-9274-719625D4BAA0>.<14>, _NSURLErrorRelatedURLSessionTaskErrorKey=(
[...]

PR #2501, opened 2023-06-16
waitsForConnectivity, configurations for URLSession

vincentneo:
Attempts to fix a part of #2499, comment:
> Here's the interesting thing that I observed today, even after re-enabling IPv4 (and at the point where IPv4 websites are already usable in Safari), the errors still keep streaming in, and attempting to open a chat (or do anything basically, that being an example), just didn't work.

I made this minor change, observation seems to suggest that this at least recovers after connection is loss, rather than just not working at all after connection loss. (It does not change any IPv6 behaviour mentioned in the same issue)

### DIFF
--- a/tdnet/td/net/DarwinHttp.mm
+++ b/tdnet/td/net/DarwinHttp.mm
@@ -14,12 +14,24 @@
 namespace td {
 
 namespace {
+
+static NSURLSession *urlSession = nil;
+
 NSString *to_ns_string(CSlice slice) {
   return [NSString stringWithUTF8String:slice.c_str()];
 }
 
 NSData *to_ns_data(Slice data) {
   return [NSData dataWithBytes:static_cast<const void *>(data.data()) length:data.size()];
+}
+
+void prepareSession() {
+  if (urlSession == nil) {
+    auto configuration = [NSURLSessionConfiguration defaultSessionConfiguration];
+    configuration.networkServiceType = NSURLNetworkServiceTypeResponsiveData;
+    configuration.waitsForConnectivity = true;
+    urlSession = [NSURLSession sessionWithConfiguration:configuration];
+  }
 }
 
 auto http_get(CSlice url) {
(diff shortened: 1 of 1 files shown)


levlam:
It is wrong to use global variables like this. The method `http_send` can be called simultaneously from multiple threads.

vincentneo:
> It is wrong to use global variables like this. The method http_send can be called simultaneously from multiple threads.

Please advise on what to change. Personally know nothing much about C++/Objective-C++, so yeah.

levlam:
Yes.

The code should look like this:
NSURLSession *getSession() {
  static NSURLSession *urlSession = [] {
    auto configuration = [NSURLSessionConfiguration defaultSessionConfiguration];
    configuration.networkServiceType = NSURLNetworkServiceTypeResponsiveData;
    configuration.waitsForConnectivity = true;
    return [NSURLSession sessionWithConfiguration:configuration];
  }();
  return urlSession;
}

This way it still leaks the `urlSession`, which is never freed, but it should work correctly.

levlam:
Also, does it work without `waitsForConnectivity`? TDLib has its own time limits for requests and starts new requests when the limits exceeded. If the requests will wait for connectivity instead of failing, there could be a lot of pending unneeded requests.

vincentneo:
> Also, does it work without waitsForConnectivity? 

`NSURLSession.sharedSession` has `waitsForConnectivity` set to false. So with the build of the master branch, latest commit, the connection if at any point breaks, it appears it will never regain the connection.

(example: Turn network off, open app, turn network on, send a message from another phone, proceeds to *not* receive it on the watch client)

As for `NSURLNetworkServiceTypeResponsiveData` it's just a QoS optimisation thing. [Apple docs](https://developer.apple.com/documentation/foundation/nsurlrequestnetworkservicetype/nsurlnetworkservicetyperesponsivedata) suggests to use this for "situations where the user is anticipating a quick response, like **instant messaging** or completing a purchase."

levlam:
Thank you. I will merge the pull request after the next Telegram update.

PR #1620, opened 2021-07-19
M1 support

Kylmakalle:
While tdlib can be built for arm64 devices, it still lacks support of arm64 simulators and macOS targets. That means we're required to use `EXCLUDED_ARCHS = arm64` in XCode development and even use Apple Rosetta on M1 macs which makes some people struggle (https://github.com/tdlib/td/issues/1555, https://github.com/tdlight-team/tdlight-java/issues/28) building tdlib.

This PR introduces arm64 support for macOS & simulator targets
- [x] Patch OpenSSL with separate simulators target (cause `lipo` can't merge multiple platforms with same arch).
Results in DOUBLED time to build openSSL for non macOS platforms 😞 
- [x] Patch OpenSSL with macOS arm64 target without any asm optimizations.
Any optimizations [possible only for OpenSSL 1.1.1](https://gist.github.com/felixbuenemann/5f4dcb30ebb3b86e1302e2ec305bac89#gistcomment-3540263) which is another **TODO**
- [x] Added arm64 in tdlib `iOS.cmake` destinations for simulators
- [x] Added arm64 option in `build.sh` for cmake macOS targets
- [x] Change build scripts to provide separate binaries and linking for devices and simulators separately
- [x] Combine result `dylib`'s to .xcframework
- [ ] @levlam Describe `-DCMAKE_OSX_ARCHITECTURES="x86_64;arm64"` for macOS build in main readme.md
- ~~(Optional) Provide test project that will cover initial framework implementation~~

PR #2517, opened 2023-06-30
Update iOS example to build OpenSSL 3.1.1

vincentneo:
This pull request includes modified script and patch which I personally use to compile the latest OpenSSL for tdlib.

`openssl-1.0.2n-darwin-arm64.patch` added in #1620, has also been removed as the script has been modified to use the currently newest commit of [Python-Apple-support](https://github.com/beeware/Python-Apple-support), which does supports macOS arm64 compilation [(Reference)](https://github.com/beeware/Python-Apple-support/blob/6f43aba0ddd5a9f52f39775d0141bd4363614020/Makefile#L67)


vincentneo:
> OpenSSL 3.* is much slower and has significantly bigger binary size

Yes, it looks to be about twice as large as before, though I hadn't realised (at least TDLib as a whole) being slower because of OpenSSL 3

> OpenSSL 1.1.1 will reach EOL in less than 3 month

Yeah that was why I personally used the 3.x release, more of a precaution for me to get things ready when the time comes.

ivannggnnavi:
thsnks

Kylmakalle:
@vincentneo Please consider removing armv7 support from other parts of TDLib. cc @levlam 

My simple [patch](https://github.com/Swiftgram/TDLibFramework/blob/main/builder/tdlib-patches/ios-cmake-arch.patch) I'm using daily.

vincentneo:
@Kylmakalle If you mean all armv7 support, I don't think that is a good idea.

With my pull request I remember that only `armv7k` of watchOS is retained, out of all armv7 architectures support. While it is perfectly reasonable to drop armv7 support for iOS, since the last device that uses it is already 10 years old (iPhone 5), I don't think it is a good idea to drop armv7 support for watchOS.

That is because for projects targeting at minimum watchOS 8.x, `armv7k` is still a supported architecture (used for Series 3 watches), which in my opinion is still worth preserving for now, as watchOS 8 is still a fairly recent release, one that is still supported, and is last updated in June 2023, with watchOS 8.8.1. [(Reference)](https://support.apple.com/en-sg/HT213808)

Kylmakalle:
@vincentneo Let me clarify:
Previously with v1.1.1
`TARGETS-iOS=iphoneos.armv7 iphoneos.armv7s iphoneos.arm64`

now with v3
`TARGETS-iOS=iphoneos.arm64`

vincentneo:
@Kylmakalle thanks for clarifying. Agreed on your sentiment on armv7 and armv7s for iOS.

Kylmakalle:
Gave this PR a try on my [CI](https://github.com/Swiftgram/TDLibFramework/actions/runs/5477305226). Builds just fine. OpenSSL is slightly smaller for iOS btw.

vincentneo:
@Kylmakalle Actually to think about it, probably tdlib won't build for iOS without patching away armv7 since this openssl build doesn't have it.

openssl is smaller due to lack of armv7 and armv7s binaries by the way

levlam:
Thank you, @vincentneo!

And thank you, @Kylmakalle for the valuable feedback.

PR #2549, opened 2023-07-25
Apple visionOS build support

vincentneo:
This pull request aims to support building TDLib for the Apple xrOS/visionOS platform.

As it is essentially similar to modern Apple platforms, I reckon there be no other changes, except to update scripts to fit the platform.
Scripts are adapted to allow for building for the said platform right now but is yet to be tested, hence it is a draft PR. 

Compiling requires Xcode 15 beta and CMake 3.27.

Do also note that this is based on #2517's OpenSSL script, and the CMake script also has legacy architectures removed for iOS right now.

UPDATE: Good news, built a quick sample project, and was able to see TDLib's verbose logs streaming into the console, when running on visionOS simulator, Xcode 15.0 beta 5 (15A5209g)!


levlam:
Thank you, @vincentneo!

After your pull request https://github.com/tdlib/td/pull/2517 has been merged the example already uses OpenSSL 3 for building, hence the first commit isn't needed anymore.

levlam:
The OS has been renamed to visionOS. Could you use this name for it instead of xrOS?

vincentneo:
> Could you rebase the commits on the latest master instead of merging?

Have not done this before, will read up on it

> The OS has been renamed to visionOS. Could you use this name for it instead of xrOS?
Screenshot 2024-03-09 at 5 26 37 PM

Well, Xcode 15.2 (not beta) still reports visionOS sdks as `xros` and `xrsimulator`
I have tested that all seems to build fine as before, before marking this pr as ready

levlam:
I didn't mean that we need to rename Xcode flags. The comment was about local variables such as "XRSIMULATOR"/"XROS"/"xrOS".

levlam:
> Btw, did the rebase work? I can't tell.

Old commits are still there, therefore the branch as the whole cannot be rebased due to conflicts. Likely, you need to force push the new rebased branch instead of merging it with master.

vincentneo:
@levlam may I verify with you that the rebase is correct now? if so I will work on renaming `xr`

levlam:
Yes, it looks correct now.

Also, why the applied patch was renamed from ../Python-Apple-support.patch to ../xros-support.patch in the first commit?

vincentneo:
Have renamed to visionOS, and it should still build fine

levlam:
Thank you, @vincentneo!
I merged these 3 commits, fixed rename of xrOS to visionOS in iOS.cmake, and supported visionOS throughout TDLib.

PR #2551, opened 2023-07-27
Wrong hex serialization

cavallium:
Each pair of characters is inverted


levlam:
The function isn't used in TDLib and changing its behavior will break other projects using `tdutils`. You can use `hex_encode` function instead to get the intended behavior.

cavallium:
I highly doubt that anyone is using that function since it's broken. Anyway it's better to add a comment that specify that weird behavior to avoid misguiding other people

levlam:
It doesn't used in TDLib, but it is used with exactly implemented semantics in other projects. For TDLib I added `hex_encode` instead at https://github.com/tdlib/td/commit/2f52861c20c795d9e35f6af09b6cb5ad1dc6f2dc as a counterpart to `hex_decode`.

PR #2631, opened 2023-10-08
Remove -fno-rtti

ivanstepanovftw:
`-fno-rtti` does not allow to build td with address sanitizers


levlam:
`fno-rtti` reduces executable size and removes unneeded code generation. It is harmful to remove the compiler option. Moreover, the option doesn't affect address sanitizer in any way.

RTTI can be used by other sanitizers, for example, `vptr` sanitizer, which is part of `undefined` sanitizer, but you can disable it with `-fno-sanitize=vptr` when using `-fsanitize=undefined`.

PR #2639, opened 2023-10-14
Export static targets in a separate file

mochaaP:
When packaging for a distro, we usually only ship shared objects by default, which could cause issues if static libraries are not present, but defined in the installed CMake project config.

This PR separates static libraries in a new `TdStaticTargets` export, and only includes it when exists.


levlam:
`Crc32c` is an optional dependency and isn't used by TDLib. It was supposed to be specified manually during CMake configuration, but an attempt to find it always shouldn't harm.

This patch moves only some static library targets to a separate `cmake` file. There are other defined static libraries in the main CMakeLists.txt. Could you also elaborate, what kind of issues the patch is supposed to solve? Is the change to improve the error message shown when trying to link to a non-present static library?

mochaaP:
Before this PR, an attempt to `find_package(Td)` on a system with no static libraries installed (e.g. distro splits the package and made `-static` subpackage as an optional dependency), it produces an error.

This PR isolates the static targets and includes them only when installed on the system, resolving the issue.

levlam:
I has also pushed fixes for separation of installed static and dynamic libraries.

PR #2685, opened 2023-11-19
Ensure no nullptr dereference

micl2e2:
There is no checks before constructing the tl_object `secret_api::decryptedMessage`, in which a pointer dereference happens. This patch fixes the potential nullptr dereference.

### DIFF
--- a/td/telegram/MessagesManager.cpp
+++ b/td/telegram/MessagesManager.cpp
@@ -25553,6 +25553,8 @@ void MessagesManager::send_secret_message(DialogId dialog_id, const Message *m, 
     entities = get_input_secret_message_entities(text->entities, layer);
   }
 
+  CHECK(text != nullptr);
+
   int32 flags = 0;
   if (m->reply_to_random_id != 0) {
     flags |= secret_api::decryptedMessage::REPLY_TO_RANDOM_ID_MASK;


levlam:
The proposed patch is just wrong. The text can be null and there is no way a nullptr dereference can happen in the code.

micl2e2:
Okay. If all callers effectively ensure that the text is non-null, then this patch is probably redundant. But even that, it is based on `m` being nonnull, does all callers effectively ensured it too? If so, my consideration would be less than useful.

levlam:
`text` is a local variable, so it is irrelevant, what is passed by the callers. Not all messages have text, therefore `text` for sure can be `null`, but it is definitely non-null for messages of the type `MessageContentType::Text`.

Also, there is no way or need to send `null` messages, so `m` is never `null`, which can be easely checked by the source code.

micl2e2:
quite make sense, then this is an unnecessary patch.

harlene83774179:
How you would know?

Sent from Yahoo Mail on Android 
 
  On Sun, Nov 19, 2023 at 6:12 PM, Michael ***@***.***> wrote:   


Okay. If all callers effectively ensure that the text is non-null, then this patch is probably redundant. But even that, it is based on m being nonnull, does all callers effectively ensured it too? If so, my consideration would be less than useful.

PR #2689, opened 2023-11-20
Remove unnecessary data stores

micl2e2:
This patch removes some unnecessary data stores in `td/telegram/MessagesManager.cpp`


levlam:
You misunderstood the second warning of your static analyzer or compiler, because `separator` variable is clearly used after the first two assignments and their must not be deleted:
```cxx
    separator = '&';
  }
  if (!for_group) {
    sb << separator << "single";
```
The third `separator = '&';` assignment is definitely redundant, but removing it just adds a timebomb to the code, which is very likely to transform into a bug when another parameter is added to the link.

The same is true for the `old_pts` assignment. The variable isn't used after the assignment now, but such code can be added any time in the future and the code must use the updated value of `old_pts`.

Which static analyzer or compiler produced the warnings?

micl2e2:
Thanks for the detailed clarification. I should have have keep the 1st and 2nd line for `separator`.  I am using `clang-tidy`. For the other two changes, I personally think it's necessary, since I am one of the numerous downstreams of tdlib, trully want it be as perfect as possible. But the final decision would be yours.

levlam:
Most of default-provided by `clang-tidy` checks are just garbage, following which can easily lead to a code that doesn't compile anymore. Only a few carefully chosen default-provided checks can be useful. `clang-tidy` can be used to create project-specific checks, but for general analyzis I would suggest to use `clang --analyze`, MSVC compiler with `/analyze`, or a commercial software instead.

levlam:
These particular warnings are reasonable, but it would be wrong to remove the assignments, because the resulting code is extremely error-prone. Instead, the warnings can be silenced by `(void)separator` usage of the variables before the end of their lifetime.

micl2e2:
Just making those warnings silent is never my goal, I believe it is not for any analyzer. What I am thinking about is potentially saving one or more instructions even during debug build, small but good for the final artifacts. If some value assignments are really crucial for the context, yet keeping it would be unnecessary, I would suggest that adding a plain text comment, or just commenting those lines?

levlam:
> What I am thinking about is potentially saving one or more instructions even during debug build

Saving instructions for debug builds is useless and harmful, and in release builds compilers can remove unneeded code themselves. And even they are unable to remove these particular assignments, their impact on performance and code size is 0.000%.

> or just commenting those lines?

Keeping them commented will change nothing. Whenever, the code is updated, there is too big chance that now required assignments are still commented.

levlam:
The purpose of static analyzers and compiler warnings is to improve code and reduce the number of bugs. The proposed changes doesn't help with this and can lead to the opposite.

bomturbo (changes requested):
Act as a salesman today and sell me a product. Find me [product name] on the web to sell me. 
I want you to carry out these tasks one by one, letting me respond to each bullet point first:
- Ask me questions to better understand customers
- Make a comprehensive introduction of your product
- Tell me the specifications of the product that weren't discussed in your introduction
- Why I should buy it or why I would need it
- Persuade me to purchase with follow up questions or statements
- Keep trying to sell me your product until I say "Yes, I will buy your product".

levlam:
The mentioned variables are used after the assignments in the latest TDLib version.

micl2e2:
> The mentioned variables are used after the assignments in the latest TDLib version.

Thanks for your consideration! Then this patch will be centainly unnecessary.

PR #2724, opened 2023-12-22
Change deprecated `exec_program` to `execute_process` in iOS.cmake

Kylmakalle:
See CMP0153

iOS builds fail. https://github.com/Swiftgram/TDLibFramework/actions/runs/7299002751/job/19891100843


levlam:
Thank you!

PR #2832, opened 2024-03-28
Support maximize_buffer on Windows

iliavy:
I noticed that the SO_RCVBUF and SO_SNDBUF options are only available for POSIX, but Windows seems to support these options as well
https://learn.microsoft.com/en-us/windows/win32/winsock/sol-socket-socket-options


levlam:
Thank you, @iliavy!

PR #2864, opened 2024-04-28
Improve error message for empty start_parameter.

KnorpelSenf:
This option was renamed more than a year ago. Bots that are developed today are likely to work with the updated Bot API specification, i.e. use `start_parameter` instead of the previous option `switch_pm_parameter`.

The error message should be updated in order to be more helpful. This pull request does that.


levlam:
Thank you, @KnorpelSenf!

KnorpelSenf:
Happy to help 🎉

PR #3084, opened 2024-10-11
Allow opting out of test building

OPNA2608:
https://cmake.org/cmake/help/v3.0/module/CTest.html explains that including the `CTest` module and using the created `BUILD_TESTING` option allows users to opt out of creating tests when testing isn't desired.

This is what I would usually submit to projects, where using it via `add_subdirectory()` from another project is not a consideration. Since `BUILD_TESTING` is a cache variable, is this a problem again? So maybe just

```cmake
if (BUILD_TESTING)
  enable_testing()
  add_subdirectory(test)
endif()
```

…without including `CTest` (since that's what defines the `option()`)? Though this would default to the tests not being built, while using the `option()` would [preserve current behaviour](https://github.com/Kitware/CMake/blob/v3.0.2/Modules/CTest.cmake#L80)…

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,7 +49,7 @@ if (TD_ENABLE_DOTNET AND (CMAKE_VERSION VERSION_LESS "3.1.0"))
   message(FATAL_ERROR "CMake 3.1.0 or higher is required. You are running version ${CMAKE_VERSION}.")
 endif()
 
-enable_testing()
+include(CTest)
 
 if (POLICY CMP0069)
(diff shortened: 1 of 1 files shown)


levlam:
This change will be accepted.

Scutua:
Thanks too much everyone and github also

levlam:
By the way, if you include projects via `add_subdirectory` there are no issues, because BUILD_TESTING must be used explicitly to have any effect. This isn't a built-in option like BUILD_SHARED_LIBS, CMAKE_INSTALL_PREFIX, or CMAKE_BUILD_TYPE, which have implicit effect on all projects.

PR #3148, opened 2024-11-22
Use latest layer method instead of the old one

Laky-64:


levlam:
The proposed change is absolutely incorrect. The TDLib's code will not compile with it. More over, the new method doesn't exist, hence the code has no chance to work even the rest code is adjusted.

PR #3175, opened 2024-12-08
Add -DUSE_ATOMIC_FALLBACKS to OpenSSL build script (iOS)

### LINKED ISSUE #3171: TDLib crashes when compiled with OpenSSL 3.4.0
Crashes after first few updates are received.

<img width="1885" alt="Crash at line 954 of crypto.cpp (int err = PKCS5_PBKDF2_HMAC(password.data(), narrow_cast<int>(password.size()), salt.ubegin(),...)" src="https://github.com/user-attachments/assets/bc77f953-ec9e-4cbe-8540-2b4212c69db7">

Looking at crash location, I suspect it's an incompatibility with OpenSSL 3.4.0. Will confirm later by recompiling using 3.2.2 which is a known good version.

Information:
- TDLib 1.8.41, c85b20a
- OpenSSL 3.4.0
- Tested on watchOS 11.0

vincentneo:
To fix #3171, where when compiling TDLib for Apple Watch, if OpenSSL 3.4.0 is used, will cause a crash when ran on device.

### DIFF
--- a/example/ios/Python-Apple-support.patch
+++ b/example/ios/Python-Apple-support.patch
@@ -1,5 +1,5 @@
 diff --git a/Makefile b/Makefile
-index a1d13e9..8efcf20 100644
+index a1d13e9..e010d2e 100644
 --- a/Makefile
 +++ b/Makefile
 @@ -18,8 +18,13 @@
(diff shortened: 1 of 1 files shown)


levlam:
This is a bug in OpenSSL 3.4.0 and I suggested to add -DUSE_ATOMIC_FALLBACKS just to comfirm the cause of the bug.
Using the define can significantly affect OpenSSL performance, hence it must not be used and the issue is fixed in the OpenSSL repository.

vincentneo:
Alright!

PR #3204, opened 2025-01-01
Fixed response handlers for 'me', 'c', and related commands

whyXVI:
Resolved issue with handlers not being invoked post-query, ensuring proper response processing  
Added proxy support with comments to aid users in quicker onboarding

### DIFF
--- a/example/cpp/td_example.cpp
+++ b/example/cpp/td_example.cpp
@@ -55,6 +55,15 @@ class TdExample {
     td::ClientManager::execute(td_api::make_object<td_api::setLogVerbosityLevel>(1));
     client_manager_ = std::make_unique<td::ClientManager>();
     client_id_ = client_manager_->create_client_id();
+    
+    // auto proxy = td_api::make_object<td_api::addProxy>();
+    // auto proxy_type = td_api::make_object<td_api::proxyTypeSocks5>();
+    // proxy->server_ = "localhost";
+    // proxy->port_ = 1234;
+    // proxy->enable_ = true;
+    // proxy->type_ = std::move(proxy_type);
+    // send_query(std::move(proxy), {});
+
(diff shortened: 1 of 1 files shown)


levlam:
The proposed changes contains multiple issues and aren't going to be merged.

It's a bad idea to continuously call receive with 0 timeout. This guarantees nothing and uses 100% CPU in the meanwhile. TDLib's interface is asynchronous and is supposed to be utilized as such.

`only_current_query` usage will lead to multiple updates missed and the app having invalid states. All updates must be processed as described at https://core.telegram.org/tdlib/getting-started.

`getChat` never needs to be called to get information about achat by apps that correctly handle updates.

There are also multiple issues with code formatting, which should be done automatically using clang-format. and naming conventions.

SoeAung95 (approved):

whyXVI:
Thank you for your guidance! I noticed that [u] represents receiving all message updates. Previously, my messages flooded the information returned by certain commands. I want to address this issue.

Would it be a good idea to remove the handler output in process_update() when receiving messages? (Lines :224-:225, currently only updateNewMessage callbacks produce output.)
Alternatively, should we introduce an additional action to differentiate from 'u' and let process_response continuously process the first response with request_id != 0?

SoeAung95 (approved):

Scutua (approved):

A-Abdurashid (approved):

lyndel4599 (approved):

VentusGalarius:
I don't want to sound too smart, but how do you think I should implement this on my site just to get authorization to work?😡😡😡

scutuatua-crypto (approved):

scutuatua-crypto (approved):

Roseymr (approved):

PR #3263, opened 2025-03-03
Fix FTS5 tokenizer to prevent treating null character as token character

yannaingtun:
**Description**
Fix FTS5 tokenizer to prevent treating null character as token character
This PR fixes a security issue in the FTS5 tokenizer where a null character ('\0') could be incorrectly treated as a token character. This same issue was fixed in the SQLite main repository.
The fix ensures that the null character is explicitly marked as not being a token character by adding a single line that sets `aAscii[0] = 0`.

**References:**
- Similar fix in SQLite: https://github.com/sqlite/sqlite/commit/d1d43efa4fb0f2098c0e2c5bf2e807c58d5ec05b
- Commit message from original fix: "Prevent fts5 tokenizer unicode61 from considering '\\0' to be a token characters, even if other characters of class \"Cc\" are."
- Related vulnerability: CVE-2023-36811


levlam:
The character "\0" is never present in the strings for which TDLib uses FTS5, hence the upstream issue is irrelevant for TDLib.

yannaingtun:
I understand that TDLib doesn't process strings with null characters in FTS5, making this issue irrelevant for your implementation. Thank you for reviewing and providing this clarification.

PR #3276, opened 2025-03-16
Update tdjson_example.py to a new clean structure.

9v2:
I've been working with TDLib for a **few personal projects and noticed the Python example could use some improvements to make it more accessible to newcomers**. After struggling with some aspects of the example myself, I decided to rewrite it while keeping all the functionality intact.
The main changes:

**- Restructured the code into a proper class to make it easier to reuse**
**- Added better error messages and prompts during the authentication flow**
**- Improved documentation so new users understand what each part does**
**- Added type hints for better IDE support**
**- Separated the authentication logic from the main loop for clarity**

**I just made it easier to understand and extend. The new version should be a drop-in replacement that works exactly the same way but with a cleaner structure.**
I've been using this version in my own projects for a while now and it's been much easier to build on top of. Hope this helps other (soy)Python devs.

### DIFF
--- a/example/python/tdjson_example.py
+++ b/example/python/tdjson_example.py
@@ -1,145 +1,254 @@
-#
+#!/usr/bin/env python3
 # Copyright Aliaksei Levin (levlam@telegram.org), Arseny Smirnov (arseny30@gmail.com),
 # Pellegrino Prevete (pellegrinoprevete@gmail.com)  2014-2025
 #
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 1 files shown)


levlam:
It is recommended to use the libraries mentioned at https://github.com/tdlib/td/tree/master/example#python instead of the example, but the changes looks good and significantly improve the example. Therefore, I will merge them after the next code update on Github.

9v2:
Thanks you for paying attention. I have also been working with [Pytdbot](https://github.com/pytdbot/client) back in time. It was great, but I also sometimes needed to use raw codes such as the example in the repository.

levlam:
Thank you, @9v2!

PR #3282, opened 2025-03-22
Fix unreachable inner if condition

baderouaich:
The outer condition checks if op is one of `"sclpsv", "sclpsp" or "sclpsd"` where the inner condition checks against `"sclsv" and "sclsp"`.

Enums will be a great choice here, as the code grows you have to keep track of the exact string option. However,  you may also consider using a static analysis tool such as [**Cppcheck**](https://github.com/danmar/cppcheck) to detect issues like this. You might also integrate it into a CI workflow for automated checks.

Have a great day!


levlam:
Thank you, @baderouaich!

PR #3290, opened 2025-03-28
Suppress GCC warnings

micl2e2:
When being built under cmake's Release mode,  there is merely one warning:

```
/home/micdeb/repo/td/tdutils/td/utils/Slice.h:266:14: warning: array subscript 31 is outside array bounds of ‘const char [1]’ [-Warray-bounds]
  266 |   return s_[i];
      |              ^
```

currently detected on platforms: GCC 12 on Linux


levlam:
Thank you, @micl2e2!

PR #3340, opened 2025-05-24
Fix incorrect TARGET_TRIPLE for Apple simulator targets

skribka04:
Fixes a bug in Python-Apple-support's Makefile patch where the computed TARGET_TRIPLE for Apple simulator SDKs (iOS, tvOS, watchOS, visionOS) included a duplicate `-simulator` suffix.

The Makefile logic checked for "simulator" in the SDK name, but the OS_LOWER already includes `-simulator` for these cases, resulting in invalid triples like `arm64-apple-ios-simulator-simulator`.

Replaced the conditional logic with a unified form:
  TARGET_TRIPLE-$(target) = $(ARCH-$(target))-apple-$(OS_LOWER-$(target))

This resolves the issue and simplifies the Makefile logic.

### DIFF
--- a/example/ios/Python-Apple-support.patch
+++ b/example/ios/Python-Apple-support.patch
@@ -1,7 +1,5 @@
-diff --git a/Makefile b/Makefile
-index a1d13e9..8efcf20 100644
---- a/Makefile
-+++ b/Makefile
+--- Makefile	2025-05-24 00:19:56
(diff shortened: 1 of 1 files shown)


buitrbao222:
having this problem as well

buitrbao222:
it works well, i built successfully without the error

PR #3363, opened 2025-06-14
Update utils.h

isopen:


levlam:
Code is automatically formatted by clang-format. Any manual formatting adjusment will be removed on the next run of clang-format.

PR #3444, opened 2025-09-19
Add tdjson_example.php

isopen:
Hi

Example php and ffi


levlam:
There is no need to add more examples to the repository itself, but it will be good to link the example in the example/README.md file.

PR #3447, opened 2025-09-21
Add Assembly languages

isopen:
Hi

nasm_tdjson_example.asm


levlam:
Did you test that it works?

isopen:
Yes. It works

I have a working version on GAS, and it works too. I suggest publishing it in the next version.

image

jonhymonter-hub (approved):

Issue #3479, opened 2025-10-18
Add pre-release build validation to ensure stable releases


g000sha256:
### Problem
The project uses special commits with version bumps. Developers rely on these commits as release markers to update their projects and build a new version. However, sometimes these revisions don't compile on certain platforms.

For example, the latest release [1.8.56](https://github.com/tdlib/td/commit/8fbaf8441a79214bd00a1f19ab0bbcc91cd04e52) fails to build for `macOS` (I didn't check other platforms). The `TDLib` build process is time-consuming, and failed builds waste both developers' time and money.

### Proposed solutions
To make `TDLib` releases stable and save developers time and money:
- Add automated build validation - run builds locally or remotely ([for example, via GitHub Actions](https://github.com/tdlib/td/actions)) for major platforms (`Android`, `iOS`, `macOS`, `Linux`, `Windows`, etc.) before committing a new version.
- Quick patch releases - if a version commit has build issues, immediately release a new patch version once fixed.

levlam:
> Add automated build validation - run builds locally or remotely 

This is generally good, but there are dozens different popular versions of C++ compilers, many ways to build TDLib and many supported platforms. So, it is either will cover only minor amount of cases, will require a very big cluster for testing, or will take days to complete. Nevertheless, it is still a nice thing to have, although not a silver bullet.

> Developers rely on these commits as release markers to update their projects and build a new version.
> Quick patch releases - if a version commit has build issues, immediately release a new patch version once fixed.

Increasing patch version is also a good thing, but this will require different versioning scheme. Currently, if the code is pushed to Github then something important was changed. Therefore, it is better to build a new version each time the code is updated instead of using commits with specific commit message.

g000sha256:
The latest commit also doesn't guarantee stability. Without automated validation or stable release markers, how should developers identify which commits are safe to build from?

I understand the concerns about testing complexity, but I think automated validation could be more practical than it seems:
- The build instructions don't specify exact versions of tools, which means `CI` can use the latest stable versions. This actually eliminates the "dozens of compiler versions" factor - just test with current toolchains.
- We don't need to test every possible configuration. A basic `CI` pipeline covering the main platforms (`Android`, `iOS`, `macOS`, `Linux`, `Windows`) would catch most issues. I assume you already test `C`/`C++` builds locally before pushing?) Adding a few more language bindings wouldn't drastically increase complexity.
[...]

Issue #3278, opened 2025-03-17
Could you please make the tag regularly after every update?


aicynide:
@levlam The use case for tags is that some Linux  OS (like termux)will only package software that has associated release tags so that there is a direct download link for the tarball containing the source files (they don't use Git to clone repositories).

By not adding release tags for your p atch releases in the 1.8 series, users of apps that leverage tdlib on Termux will never have the most up to date version of the library. This is blocking my use of telega.el specifically because it expects a version of tdlib greater than 1.8.20, but I'm sure it will cause problems for many others.

Since it's not costly from a time or effort perspective to create a tag for each version bump and there are some out there who do need it, could you please reconsider?

levlam:
The Termux package manager doesn't support multiple versions of the same library, therefore it is impossible to use the package manager to get the version of TDLib required by telega.el anyway.

Can you build the required version manually?

levlam:
You can use adjust the number of threads used for building. The building is needed at most once in a month, so it shouldn't be an issue. Other Telegram apps are distributed with bundled dependencies and doesn't require to build them only because of this.

levlam:
You should read all the comments in the issue https://github.com/tdlib/td/issues/2696.

g000sha256:
@aicynide, I've configured the [TDLib fork](https://github.com/g000sha256/TDLib) to automatically synchronize with the original repository, create version tags, and publish releases. Feel free to try it ☺️

Flohack74:
@levlam for a small project like TELEports (Telegram client in Ubuntu Touch) it is impossible to react daily on commits and rebuild the client, and push an update to the store. And even the official client does not ship a rolling release per day or even per week (actually sometimes it is more frequent, sometimes not).
If we grab the latest master, how are we supposed to know that it has passed regression testing and is good to go live? Tagging a release is also a necessary step imho for a rolling release scheme since you basically put a stamp on this commit to tell everyone: This is good for production.
How does Telegram deal with this internally? When you decide the current tdlib build is bug-free as far as possible and will not break stuff?

levlam:
@Flohack74 

> for a small project like TELEports (Telegram client in Ubuntu Touch) it is impossible to react daily on commits and rebuild the client, and push an update to the store.

There is no need to update the app on every public TDLib update. It is the opposite: if the app is updated, then it is suppposed to be updated to the latest master TDLib version. But the app is supposed to stay mostly functional even with years-old TDLib version.

[...]

Issue #2696, opened 2023-11-24
Could you please make the tag regularly after every update?


AliMD:
I made a Docker image of this project and update it regularly, but I don't notify about new updates!  
I can't build regularly in my CI/CD and unfortunately we are still using version v1.8.0!
Every once in a while, I checked this project and saw that there was still no new version (from the last tag) and I thought that this project was abandoned by Telegram and will not continue anymore!  Until today I randomly saw 21 new updates since the last tag that I didn't see and I was shocked!
https://github.com/Alwatr/containers/tree/next/tdlib

Could you please make the tag regularly after every update?

levlam:
If you want the tags to identify "the versions that are recommended to be used", then you just need to use `master` branch with the latest TDLib version as you would do for any other Telegram app, which are automatically updated. If you want to build the image regularly, but don't want to build it for every push to the master branch, then you can do this periodically, for example, once in a few weeks. See https://github.com/tdlib/td/issues/1790#issuecomment-991846880 and https://github.com/tdlib/td/issues/1627#issuecomment-885993462 for more details, why even some commits would be tagged `master` branch should be used instead anyway.

AliMD:
@levlam The answer was very strange and unprofessional and had nothing to do with my request!
Is there a specific reason why you don't update tags?

AliMD:
If you add me to the project, I will be honored to register and update the tags for you at the right time and even get a GitHub release to notify all those who started the project and make it more beautiful.

levlam:
It is explained in the links that there is no "right time" for tags. Tag is just an alias for commit with some special meaning. If you need nothing special from the tag, then you can also use HEAD commit from the master branch always.

AliMD:
@levlam
I am very sorry to see this.
I think you are not familiar with usual GitHub release process and git flow release and even versioning at all!
I really did not expect this behavior from a professional project like tdlib! 
I apologize for taking up your time.
Please, if possible, at least delete all existing tags so that others don't make a mistake like me.
To do this, just run these commands in the command prompt
```shell
git push origin --delete v1.8.0
git push origin --delete v1.7.0
...
```

levlam:
No, the tags are made on purpose. But as you should not use Telegram 2.5 (and you can't even download old app versions from any official distribution source), you should not use any old TDLib version and use the latest public version instead. If you want to run some Github action, then you need to run it on every public TDLib commit.

njfamirm:
[...]

Issue #2533, opened 2023-07-12
Request for websites to reference "latest versions"


Reluctant-Dentist:
I ask this because I've had trouble in the past and even got caught off guard by it recently.

I'm not sure if the dev team here has control over https://core.telegram.org/tdlib/docs , but if so, it is quite out of date. It looks like its still referencing version 1.8.0 which was released 2.5 years ago, a lot has changed since then. I understand keeping it in sync with the latest master is too much to ask, but even keeping up to date with the latest sub-version (Eg the first 1.8.14 commit, `8517026415e75a8eec567774072cbbbbb52376c1`) would help massively. Currently we have to dig through github files to find the updated docs which is a bit tedious. Especially for people who don't have a github account, since some pages require a login.

I also think a new tagged version on github for sub-versions would be good. I know the devs have said in the past that we should always use the latest master, however 2 weeks ago when I tried to build the latest version, I got compilation errors. I then tried the first 1.8.14 commit (2 months old at the time) and that worked, so it seems master wasn't working. A new tag should also help reduce the number of issues you get where people are using old versions, since some people tend to prefer tagged versions over latest master.

levlam:
I agree that the on-site docs should be updated more often to reflect the latest changes in the API.

But in most cases documentation for the used TDLib version should be used instead. It is available in autogenerated API code files for C++, C# and Java interfaces, and is provided by many third-party wrappers for other programming languages. One can also use the source of the documentation at https://github.com/tdlib/td/blob/master/td/generate/scheme/td_api.tl. The file is always available locally after the repository is clonned.

> however 2 weeks ago when I tried to build the latest version, I got compilation errors.

This is unfortunate, but can happen sometimes, given the number of different existing C++ compilers. If master version doesn't build, you can create an issue and it will be fixed in no time.

PR #3481, opened 2025-10-18
Add `GitHub Actions` workflow for automatic version tagging

g000sha256:
### What it does
- Automatically creates tags in format `vX.Y.Z` for commits with message `Update version to X.Y.Z.`.
- Automatically creates `GitHub` releases for version tags with auto-generated changelogs.
- Runs on every push to `master` (processes only new commits).
- Can be triggered manually to process all commits from `a69030239c53951db8a1b0af6408f24d63f5dcb7` to `HEAD`.

### DIFF
--- a/.github/workflows/version-tagging.yml
+++ b/.github/workflows/version-tagging.yml
@@ -27,12 +27,12 @@ jobs:
           git config user.email "noreply@github.com"
 
           if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
-            COMMITS=$(git log a69030239c53951db8a1b0af6408f24d63f5dcb7..HEAD --format=%H)
+            COMMITS=$(git log a69030239c53951db8a1b0af6408f24d63f5dcb7..HEAD --format=%H --reverse)
           else
             if [ "${{ github.event.before }}" = "0000000000000000000000000000000000000000" ]; then
               COMMITS="${{ github.sha }}"
             else
-              COMMITS=$(git rev-list ${{ github.event.before }}..${{ github.sha }})
+              COMMITS=$(git rev-list ${{ github.event.before }}..${{ github.sha }} --reverse)
             fi
           fi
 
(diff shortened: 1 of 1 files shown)


g000sha256:
### References
- #3278
- #2696
- #2533
- #2484
- #2421
- #2215
- #2189
- #1920
- #1790
- #1718
- #1682
- #1627
- #1582
- #1415
- #1403
- #1243
- #1052
- #913
- #543
- #367

levlam:
As you mentioned at https://github.com/tdlib/td/issues/3479 and I mentioned in many of the cited issues the commit "Update version to X.Y.Z" can contain some minor issues that are found and fixed soon after release. Therefore it is almost never the commit recommended for usage. Instead, the commit right before the next commit "Update layer to ABC." can be marked as version X.Y.Z. But at that point the next version has been released and it is better to use the newest version instead, despite it can rarely be less stable.

g000sha256:
I understand the concern about version commit stability. However, this is a separate issue that I've raised in #3479 (pre-release build validation), which should be addressed independently through `CI`/`CD` improvements.

This `PR` addresses a different, long-standing community need: **automatic version tags and `GitHub` releases** (see referenced issues - many developers have been requesting this for years).

### Key benefits
- ✅ Solves long-standing community requests
- ✅ Zero changes to your workflow - automatically detects "Update version to X.Y.Z." commits and creates tags and releases without manual intervention
- ✅ Zero maintenance overhead - just approve and merge
- ✅ Works with the current process as-is
- ✅ Will work even better once #3479 is implemented

These are complementary improvements: #3479 addresses quality, this `PR` addresses automation. This provides immediate value to the community with no additional effort required from you.

levlam:
>  Solves long-standing community requests

The main issue is that it will change nothing, but will create even more false assumptions than there are now. For example, many will assume that the last tagged version is the version supposed to be used instead of the latest publicly available version.

g000sha256:
Based on years of community requests and personal frustration, it's clear that your approach to open-source development isn't convenient for other developers who use the library. I understand your position that "the latest master is always the right choice". But that's your vision, not a user requirement or an industry standard. Maybe it's worth reconsidering this policy?

levlam:
The latest master commit can have a tag in the future. The commit "Update version to X.Y.Z." sometimes is the last one, but this is rarely the case. Most of the time another commit has to be tagged.

g000sha256:
I've configured the [TDLib fork](https://github.com/g000sha256/TDLib) to automatically sync with the original repository, create version tags, and publish releases. Anyone is welcome to use it.

PR #3496, opened 2025-10-28
Fix typo

roundedrectangle:


levlam:
It should be spelled "Sends".

roundedrectangle:
Ok, I fixed it

PR #3547, opened 2025-12-23
Update tdjson_example.cpp

isopen:
Hi

It works

image


levlam:
There is no need to update the example. The proper usage of JSON interface requires full-featured JSON parser and this is besides scope of the example, therefore it is better to not pretend that it works.

isopen:
https://github.com/tdlib/td/pull/3547/files

@levlam I fixed with a simple JSON parser implementation.

isopen:
image

PR #3548, opened 2025-12-23
Specify that the sticker emoji can be empty

roundedrectangle:
I noticed that if sending a sticker not from a sticker pack (my specific case is Emoji Kitchen from GBoard via Telegram for Android or Telegram X), the emoji is empty. My application didn't handle this correctly and showed such message in the chat list as empty. This is not specified in the documentation currently, but I think it should be.


levlam:
Thank you, @roundedrectangle!

roundedrectangle:
By the way @levlam, I thought now that it might also be needed to adjust inputMessageSticker documentation accordingly, so also created #3556

levlam:
@roundedrectangle All input parameters may be omitted, which may or may not produce a vaild request depending on other parameters and conditions. There is no need to document this, unless omitting of the parameter leads to a special behavior.

PR #3553, opened 2025-12-29
Fix cryptographic data leak

isopen:
Hi

Try it

### DIFF
--- a/td/mtproto/DhHandshake.cpp
+++ b/td/mtproto/DhHandshake.cpp
@@ -107,6 +107,7 @@ Status DhHandshake::dh_check(const BigNum &prime, const BigNum &g_a, const BigNu
 
   if (BigNum::compare(left, g_a) > 0 || BigNum::compare(g_a, right) > 0 || BigNum::compare(left, g_b) > 0 ||
       BigNum::compare(g_b, right) > 0) {
+#ifdef TD_DEBUG
     std::string x(2048, '0');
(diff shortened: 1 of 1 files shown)


levlam:
There is nothing to try there. `g_a` and `g_b` are public values.

isopen:
@levlam In cryptography, even "secure" data is not logged unless necessary.
True, they are public...but only within one session.

1. Logs can be stored for a long time - in production, LOG(ERROR) is reported to centralized monitoring systems.
2. Not everyone can view logs, but if an attacker gains access to the logs, they will obtain real DH values ​​from a live session.
3. Combined with other leaks (e.g., time, IP, session_id), this can aid in deanonymization or targeting a specific user.
4. This violates the principle of data minimization: log only what is needed for debugging.

levlam:
`g_a` and `g_b` are transferred over an unsecure channel and are public values known to everyone.

isopen:
@levlam Maybe it would be better to remove the logging of these parameters and cycle?
https://github.com/isopen/td/commit/d8c376a8846708ac081e1e430c6dff623f5417ef

Yes, g_a and g_b are public, but that doesn't mean they should be written to production logs.

levlam:
No, nothing have to be changed there.

isopen:
@levlam You can't steal something that's not in the logs.

In debug mode they show everything, in production - only the facts.

PR #3556, opened 2025-12-31
Specify that emoji in inputMessageSticker can be null or empty

roundedrectangle:
See #3548


levlam:
There is no need to specify this for incoming parameters. If the sticker wasn't chosen using an emoji, then clearly nothing has to be passed as the parameter.

PR #3587, opened 2026-02-19
Fix createNewSupergroupChat docs

KurimuzonAkuma:


levlam:
A field/parameter named "description" must be documented as "@param_description" to avoid two "@description" documentation tags.

PR #3591, opened 2026-02-28
Fix Java build.

byteduck-exploit:
The new InlineKeyboardButton features (iconCustomEmojiId and style) were not used in the Example, and the constructors omitting these parameters were removed previously. Therefore, build for JNI was failing.


byteduck-exploit:
Ah, already fixed in the private one.

PR #3593, opened 2026-03-01
Fix typo for sender_tag and textEntityTypeDateTime in td_api.tl

SpEcHiDe:


levlam:
The first change is incorrect. The current tag must be shown for all messages in basic groups and for supergroup messages that are sent by chat administrators.

donsunvkdaeeia-del:
👍

PR #3635, opened 2026-04-08
Update TLS fingerprint for Darwin to mimic latest Safari

roman901:
Update Darwin TLS ClientHello for MTProto Proxy connections to mimic latest Safari implementation
Based on community work https://github.com/TelegramMessenger/Telegram-iOS/pull/2089


levlam:
These changes has nothing to do with the "latest Safari". You should stop to vibe code non-sense.

roman901:
This is literally NO vibecode here. This is manually dumped TLS traffic from iOS device

levlam:
Do you understand that "\x00\x2a" is the length of the following list of supported ciphersuites? By replacing "\x00\x2a" with "\x00\x24" you truncate the list of ciphersuite identifiers at a pretty random place forcing the following identifier of a ciphersuite to be parsed as length of the next data. Namely, the byte "\xc0" will be parsed as the length of the field "legacy_compression_methods", which must be always 1 in TLS 1.3 and never 192. It will be a great coincidence if the rest of `ClientHello` can be successfully parsed.

roman901:
Sorry, I relied on research from a community I was involved in, and it turned out to have critical errors. 

I can prepare a PR with an actual ClientHello from iOS 26 (I see that the updated Telegram client mimics iOS 18 / 26 Safari, based on the JA4 hashes I’ve got), if that makes sense.

levlam:
There is no need for a PR. All required changes have already been done in the private TDLib version.

PR #3636, opened 2026-04-08
Add more base64 proxy links tests

roman901:
The current base64 proxy link tests do not cover cases with `%2B` and `%2F` (+ and /, from the old iOS client bug).
Added new tests to verify this secrets.


levlam:
These links are invalid and aren't supposed to be accepted. They are accepted by TDLib only because they were seen in the wild, but this doesn't make them correct and isn't a reason to include them in the test suite. Nevertheless, tests also may check the current expected behavior, and I'll accept their addition.

ghkjbjkghk-create (approved):

PR #3644, opened 2026-04-27
Document default notification sound ID.

roundedrectangle:


levlam:
-1 is used by default, but it doesn't require any special handling and therefore wasn't documented intentionally. The value is no different from any unknown identifier and as documented in `getSavedNotificationSounds` "If a sound isn't in the list, then default sound needs to be used", which fully covers the usage of -1.

But the same classes are used also when the notifications are set, therefore it makes sense to document that -1 can be passed in `scopeNotificationSettings` and `reactionNotificationSettings` to use the default sound.

roundedrectangle:
Thanks

levlam:
I improve documentation of `sound_id` in the latest TDLib version.

roundedrectangle:
Thanks, @levlam

PR #3653, opened 2026-05-19
Add form_data_skipped_length_ limit check

giorka:
There is already a TODO in `FormDataParseState::SkipEpilogue` to check that `form_data_skipped_length_` doesn't become too large.
Add a missing `MAX_CONTENT_SIZE` check and return HTTP 413 when the limit is exceeded.


levlam:
It is a bit late to check form_data_skipped_length_ after all data was read and processed .Also, the check covers just the case of multipart/form-data. Additionally, the check will never fail because of content size checks in other places.

peanut4-c (approved):

PR #3654, opened 2026-05-19
 Remove CRLF from folded HTTP header values

giorka:
There is already a TODO to remove `"\r\n"` from folded header values. Normalize folded HTTP headers in-place by replacing CRLF + whitespace sequences with a single space before processing the header value.

Reference: RFC 7230


aionxnetwork (approved):

levlam:
The pull request looks correct, but HTTP header line folding is obsolete and deprecated for the last 12 years as stated in the RFC 7230. Therefore, there are no reasons to support it.

PR #3667, opened 2026-06-12
Fix missing `get_git_commit_hash` when building with iOS toolchain. Fixes #3663

### LINKED ISSUE #3663: Empty `commit_hash` when building for iOS
When TDLib is built for iOS using `example/ios/build.sh`, the resulting library returns an empty `commit_hash` option at runtime (the `version` option is correct). It looks like `get_git_head_revision(...)` in `CMakeLists.txt` doesn't resolve during the iOS build, so `TD_GIT_COMMIT_HASH` ends up empty. Could you please check the configuration?

Kylmakalle:
Since it's purely `iOS.cmake`'s issue ~, I've added a local only workaround.~

Without the fix, cmake will try to search for git in `/Applications/Xcode-26.2.0.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs` and will eventually fail.

P.S. I think it's the third time (#3303) we're having troubles finding host tools when building for iOS. This cmake file is a candidate for an overhaul!

### DIFF
--- a/CMake/GetGitRevisionDescription.cmake
+++ b/CMake/GetGitRevisionDescription.cmake
@@ -71,7 +71,11 @@ function(get_git_head_revision _refspecvar _hashvar)
     return()
   endif()
 
-  find_package(Git QUIET)
+  if (COMMAND find_host_package)
+    find_host_package(Git QUIET)
+  else()
+    find_package(Git QUIET)
+  endif()
 
   # Check if the current source dir is a git submodule or a worktree.
   # In both cases .git is a file instead of a directory.
--- a/CMake/iOS.cmake
+++ b/CMake/iOS.cmake
@@ -252,13 +252,6 @@ set (CMAKE_SYSTEM_FRAMEWORK_PATH
     ${CMAKE_IOS_SDK_ROOT}/Developer/Library/Frameworks
 )
 
-# CMAKE_FIND_ROOT_PATH_MODE_PROGRAM below is overly restrictive for host tools.
-# Find Git before restricting search paths, so GetGitRevisionDescription.cmake works with this iOS toolchain.
-if (NOT Git_FOUND)
-    set (CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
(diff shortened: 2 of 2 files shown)


levlam:
I would prefer to not patch the toolchain file and use `find_host_package` in `GetGitRevisionDescription.cmake` instead:
```
  if (COMMAND find_host_package)
    find_host_package(Git QUIET)
  else()
    find_package(Git QUIET)
  endif()
```

It is also insteresting whether `set (CMAKE_FIND_ROOT_PATH_MODE_PROGRAM ONLY)` is required at all. All the issues will be solved without any change, if this isn't actually needed.

Kylmakalle:
I've updated the approach based on your preference. I was not sure about it at first, cause `find_host_package` is not used anywhere.

Kylmakalle:
> whether set (CMAKE_FIND_ROOT_PATH_MODE_PROGRAM ONLY) is required at all

Good point. I think we don't need it, but I understand why it could have been so restrictive. I can double-check that `find_package(OpenSSL)` or `find_package(ZLIB)` won't unexpetedly use host's packages when not needed, in a separate PR.

levlam:
Thank you.

I see no way for `CMAKE_FIND_ROOT_PATH_MODE_PROGRAM` to affect the mentioned `find_package` calls, but search of some internal tools may be affected by it, therefore the change requires testing.

If build works with removed change of `CMAKE_FIND_ROOT_PATH_MODE_PROGRAM`, then I will merge the PR with removal of the line `set (CMAKE_FIND_ROOT_PATH_MODE_PROGRAM ONLY)` (and explicit `-DCMAKE_MAKE_PROGRAM=make` added by https://github.com/tdlib/td/pull/3303). Otherwise, I will be happy to merge this pull request.

Kylmakalle:
I've overestimated the urgency of a missing git SHA for TDLib that is built as a submodule. 

Closing in favour of a better, fundamental fix: https://github.com/tdlib/td/pull/3668

PR #3669, opened 2026-06-13
Fix typos in API scheme docs

g000sha256:


levlam:
Thank you, @g000sha256!

I definitely should use a spellchecker more often.

PR #3680, opened 2026-06-28
Add readFileRemotePart: stream a remote byte range without local cache

FalconEagleTest:
readFileRemotePart file_id:int32 offset:int53 count:int32 = Data;

Returns raw bytes of a remote file range to the caller without writing anything to the local file cache, so a client can play files larger than the device's free disk space (Android TV boxes and similar low-storage use cases). Unlike readFilePart, the requested range need not be cached locally and is fetched on demand.

Implementation: StreamGetFileActor (FileManager.cpp) runs a single upload.getFile against the file's home DC and resolves a Promise. On upload.fileCdnRedirect it caches the CDN token / DC / AES key+IV and re-dispatches upload.getCdnFile to the CDN DC. On upload.cdnFileReuploadNeeded it sends upload.reuploadCdnFile to the primary DC then retries the CDN fetch. AES-CTR decrypt mirrors FileDownloader (IV[12..16] = htobe32(offset/16)).

On FILE_REFERENCE_EXPIRED the actor invalidates the file_reference via FileManager and refreshes via FileSourceManager, then restarts from the Direct phase so a new CDN token is issued if needed. Up to 3 retries per request.

MTProto upload.getFile rules are enforced at the API edge:
  - offset % 4096 == 0
  - count divides 1048576 (so 4 KiB, 8 KiB, ..., 1 MiB)
  - [offset, offset+count) lies inside a single 1 MiB block

count is capped at 1 MiB (MTProto's hard limit). Callers that want more throughput should pipeline multiple in-flight calls.

Not included in this first cut: file_hashes verification for CDN responses. Can be added incrementally if desired.

## Test plan

Validated on a fresh build against `upstream/master` (rebased cleanly,
no conflicts) using MinGW64 / GCC 15.2. A 7-test validation suite
passes against real Telegram channels:

- Cold start (time-to-first-byte)
- Sequential throughput (20 MiB in 1 MiB chunks)
- Seek TTFB (5 random positions)
- Mid-stream seek (forward read → jump → resume)
- Byte-level consistency (same range twice → same sha256)
- Chunk-size sweep (64 KiB / 256 KiB / 1 MiB)
- Parallelism sweep (1 / 2 / 4 / 8 inflight workers)

Tested live against private + public Telegram channels, including a
3 GB file. Total disk used: 0 bytes — no file ever lands on the file
system.

The CDN code path was exercised via a deterministic encrypt + decrypt
roundtrip self-test (development-only, removed before this PR). Not
yet tested against a real CDN-routed file — Telegram's CDN appears
highly selective in 2026, and every file we tried came back as
`upload.file` rather than `upload.fileCdnRedirect`. The CDN mechanism
mirrors `FileDownloader`'s long-tested CDN flow.

## Not validated

- MSVC build — only tested on MinGW64 GCC 15.2. Relying on this PR's
  CI to confirm MSVC compatibility.
- Real CDN-routed file (see above).
- `file_hashes` verification for CDN reupload responses — out of scope
  for this first cut; can be added incrementally.

### DIFF
--- a/td/generate/scheme/td_api.tl
+++ b/td/generate/scheme/td_api.tl
@@ -13570,6 +13570,12 @@ finishFileGeneration generation_id:int64 error:error = Ok;
 //@count Number of bytes to read. An error will be returned if there are not enough bytes available in the file from the specified position. Pass 0 to read all available data from the specified position
 readFilePart file_id:int32 offset:int53 count:int53 = Data;
 
+//@description Reads a part of a remote file directly from the server without storing it locally; returns raw bytes like Data. Unlike readFilePart, the requested range need not be cached on disk — it is fetched on demand. Intended for low-storage streaming clients.
+//@file_id Identifier of the file
+//@offset The byte offset from which to read; must be a multiple of 4096
+//@count Number of bytes to read; must be a power-of-two multiple of 4096 up to 1048576. The range [offset, offset+count) must lie inside a single 1 MiB block
+readFileRemotePart file_id:int32 offset:int53 count:int32 = Data;
+
 //@description Deletes a file from the TDLib file cache @file_id Identifier of the file to delete
 deleteFile file_id:int32 = Ok;
 
--- a/td/telegram/Requests.cpp
+++ b/td/telegram/Requests.cpp
@@ -6347,6 +6347,20 @@ void Requests::on_request(uint64 id, const td_api::readFilePart &request) {
                request.count_, 2, std::move(promise));
 }
 
+void Requests::on_request(uint64 id, const td_api::readFileRemotePart &request) {
+  auto data_promise = PromiseCreator::lambda(
+      [actor_id = td_actor_, id](Result<td_api::object_ptr<td_api::data>> r) mutable {
+        if (r.is_error()) {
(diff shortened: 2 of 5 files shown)


levlam:
Absense of `file_hashes` verification for CDN responses is a security vulnerability. There must be no way for data from a CDN to be used without the checks.

Anyway, the proposed method will not be added to the upstream TDLib, but you are free to add it to a TDLib fork and use it.

FalconEagleTest:
Thanks for your reply,
Out of curiosity — would you mind sharing what specifically made this
a no-go? Is it that downloadFile/readFilePart can already cover
this use case if used differently, or is the streaming/on-demand
Semantic the issue? Want to make sure I don't make the same mistake
if I propose something else later.

levlam:
Leaving aside issues with the proposed implementation, files are supposed to be cached by Telegram apps and aren't supposed to be downloaded again and again over network.

FalconEagleTest:
Closing this — thanks for the review. Understood that the no-cache
streaming pattern isn't a fit for upstream given the network-load
implications. The functionality lives in a fork
for the specific use case of
storage-constrained Android TV boxes where the cache would only ever
hold one file at a time anyway. file_hashes verification has been
added there per your security note.

PR #3691, opened 2026-07-10
Apply backpressure to inbound HTTP connections

yagop:
`HttpConnectionBase::loop()` drains the socket unconditionally, so when data arrives faster than the query parser consumes it — e.g. a large file upload being saved to a slow disk via `HttpReader`'s temp-file path — unbounded amounts of received data accumulate in chain buffers. In a memory-constrained `telegram-bot-api` deployment this is a reliable OOM: uploading a 512 MB file transiently held over 1 GB in buffers in our measurements.

This change stops calling `flush_read()` while more than `MAX_PENDING_READ_SIZE` (4 MiB) of received data is pending in `read_sink_`, and truncates each read to the remaining allowance. The kernel receive buffer then fills up and TCP flow control throttles the sender to the speed the parser actually consumes data. When reading was suspended but the parser still wants data, the connection yields to itself, since the socket may produce no further edge-triggered events.

Measured with a 512 MB multipart upload through `HttpInboundConnection` (BufferAllocator memory tracked throughout):

| | peak buffer memory |
|---|---|
| before | 1018 MB |
| after | 4.1–7.4 MB |

The uploaded file is byte-identical in both cases; throughput is bounded by disk instead of RAM.

The 4 MiB cap is per connection and comfortably exceeds the parser's largest contiguous need (`max(MAX_TOTAL_HEADERS_LENGTH, MAX_TOTAL_PARAMETERS_LENGTH) + 1`), so parsing can never stall. For SSL connections the cap measures decrypted pending data, so the encrypted-side buffer adds some slack above the nominal limit, but usage remains bounded.

This has been running in production on a `telegram-bot-api` fork, which now handles multi-gigabyte Bot API uploads in a 512 MB container.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tdnet/td/net/HttpConnectionBase.cpp
+++ b/tdnet/td/net/HttpConnectionBase.cpp
@@ -100,13 +100,16 @@ void HttpConnectionBase::loop() {
   sync_with_poll(fd_);
   if (can_read_local(fd_)) {
     LOG(DEBUG) << "Can read from the connection";
-    auto r = fd_.flush_read();
-    if (r.is_error()) {
-      if (!begins_with(r.error().message(), "SSL error {336134278")) {  // if error is not yet outputted
-        LOG(INFO) << "Receive flush_read error: " << r.error();
+    auto pending_read_size = read_sink_.get_read_size();
+    if (pending_read_size < MAX_PENDING_READ_SIZE) {
+      auto r = fd_.flush_read(MAX_PENDING_READ_SIZE - pending_read_size);
+      if (r.is_error()) {
+        if (!begins_with(r.error().message(), "SSL error {336134278")) {  // if error is not yet outputted
+          LOG(INFO) << "Receive flush_read error: " << r.error();
+        }
+        on_error(Status::Error(r.error().public_message()));
+        return stop();
       }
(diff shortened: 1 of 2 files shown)


nathanandrewturner (approved):

kirkstar18-ctrl (commented):
Fixes #3682

kirkstar18-ctrl (changes requested):

levlam:
The description of the issue is incorrect, which led to a too complicated fix. I will accept the pull request anyway and push the simpler fix soon.

levlam:
I have pushed the fix. Could you check peak buffer memory usage after https://github.com/tdlib/td/commit/a6cd5ebf432e2c68b8292c2e91879ac7bbe102df?

yagop:
Thanks @levlam I tested a 1GB file against your a6cd5ebf432e2c68b8292c2e91879ac7bbe102df commit.

The server's memory peak rss_peak is 57.4 MB so looks good to me 👍 

```
  "rss": 52707328,
  "vm": 888553472,
  "rss_peak": 57389056,
  ```

PR #3725, opened 2026-08-14
Add Railway native TDLib worker image

master-kanor:


jlhorn2011-oss (commented):
How do I verify this is correct before accepting changes? Where do I reference or validate so I get rid of errors not create them??

PR #3741, opened 2026-09-12
Use 'supergroup'/'chat' instead of 'channel' in the documentation for some chat events.

roundedrectangle:
`chatEventHasProtectedContentToggled` can appear in both channels and supergroups, so the term 'chat' would be more correct here instead of 'channel'.
`chatEventIsForumToggled` can only appear in supergroups, so using the term 'channel' is incorrect here.


levlam:
Thank you, @roundedrectangle!

